Issue #11, opened 2022-10-21
Istio CRD proposal


strekm:
**Description**

Propose specification of Istio CRD based on following IstioOperator configuration:

```
  meshConfig:
    accessLogEncoding: JSON
    accessLogFile: /dev/stdout
    defaultConfig:
      gatewayTopology:
        numTrustedProxies: 1
```
Additionally for each Istio component hpa and resource configuration should be possible:
```
  components:
    ingressGateways:
    - enabled: true
        hpaSpec:
          maxReplicas: 5
          metrics:
          - resource:
              name: cpu
              targetAverageUtilization: 80
            type: Resource
          - resource:
              name: memory
              targetAverageUtilization: 80
            type: Resource
          minReplicas: 1
        resources:
          limits:
            cpu: 2000m
            memory: 1024Mi
          requests:
            cpu: 10m
            memory: 128Mi
```

**Reasons**

Support for xff

**Attachments**

part of: https://github.com/kyma-project/kyma/issues/15865

dariusztutaj:
Proposal:
```
apiVersion: operator.kyma-project.io/v1alpha1
kind: Istio
metadata:
spec:
  controlplane:
    istiod:
      deployment:
        hpa: 
          maxReplicas: 5
          minReplicas: 2
        strategy:
          rollingUpdate:
            maxSurge: 50%
            maxUnavailable: "0"
        resources:
          limits:
            cpu: 250m
            memory: 384Mi
          requests:
            cpu: 10m
            memory: 128Mi
      meshConfig:
        gatewayTopology:
          numTrustedProxies: <VALUE>
          forwardClientCertDetails: SANITIZE|FORWARD_ONLY|APPEND_FORWARD|SANITIZE_SET|ALWAYS_FORWARD_ONLY
  dataplane:
    ingressGateway:
      deployment:
        hpa: 
          maxReplicas: 5
          minReplicas: 2
          metrics:
          - resource:
              name: cpu
              targetAverageUtilization: 80
            type: Resource
          - resource:
              name: memory
              targetAverageUtilization: 80
        strategy:
          rollingUpdate:
            maxSurge: 50%
            maxUnavailable: "0"
        resources:
          limits:
            cpu: 2000m
            memory: 1024Mi
          requests:
            cpu: 100m
            memory: 128Mi
        podAffinity:
          podAntiAffinity:
            preferredDuringSchedulingIgnoredDuringExecution:
            - weight: 100
              podAffinityTerm:
[...]

PR #12, opened 2022-10-21
istio-manager API proposal

strekm:
**Description**

- istio-manager API proposal

**Related issue(s)**
https://github.com/kyma-project/istio/issues/11

### DIFF
--- a/docs/xff-proposal.md
+++ b/docs/xff-proposal.md
@@ -1,8 +1,16 @@
+# Istio v1alpha1 proposal
+
+## Summary 
+
 This document proposes  istio-manager API exposed to users. Initial version covers hpa and resources configuration for Istio components such as istiod, ingress gateway and sidecars. This version also covers configuration needed to forward client information to workloads (client address via xff).
+
+## Details
 
 API specification on root level was split into `controlPlane` and `dataPlane` as described in Istio overview. `controlPlane` holds resources and hpa configuration for `istiod` component and general configuration called `meshConfig`. For this moment `meshConfig` will expose configuration to define `numTrustedProxies`, which is required for xff. `dataPlane` configuration contains resources and hpa configuration for `ingressGateway` component.
 
 Status will be implemented per `kyma-operator` requirements.
+
+## Istio CR example
 
 ```
 apiVersion: operator.kyma-project.io/v1alpha1
@@ -57,6 +65,8 @@ spec:
 status:
   state: Ready | Processing | Error | Deleting
 ```
+
+## Future plans
 
 In the future `meshConfig` specification can be enhanced with tracing specific configuration by defining `extensionProvider`. `ingressGateway` specification can be enhanced to enable users to define gardener domain annotations simplifying use of custom domain to expose workload.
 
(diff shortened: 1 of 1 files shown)


dariusztutaj (dismissed):
/meow

docs/xff-proposal.md
```diff
@@ -0,0 +1,75 @@
...
+            maxSurge: 50%
+            maxUnavailable: "0"
+        resources:
+          limits:
+            cpu: 250m
+            memory: 384Mi
+          requests:
+            cpu: 10m
+            memory: 128Mi
+      meshConfig:
```
pbochynski, line 37: Why mesh config is under istiod?

pbochynski (changes requested):
Do you plan to support all [kubernetes settings](https://istio.io/latest/docs/reference/config/istio.operator.v1alpha1/#KubernetesResourcesSpec)?
It is not clear.
The API design suggests that it is a 1:1 copy of https://istio.io/latest/docs/reference/config/istio.operator.v1alpha1/#KubernetesResourcesSpec
Please make a decision: do you want to keep it consistent with IstioOperator spec or not.
If you decide to go with spec easily mapped to IstioOperator spec list all supported fields and make it more similar. If you go other way probably it would be better to have API more concise (remove dataplane, controlplane, deployment keywords that do not provide value)

docs/xff-proposal.md
strekm: afaik there's CM for `istiod` created out of `meshConfig`.

strekm:
> Do you plan to support all [kubernetes settings](https://istio.io/latest/docs/reference/config/istio.operator.v1alpha1/#KubernetesResourcesSpec)? It is not clear. The API design suggests that it is a 1:1 copy of https://istio.io/latest/docs/reference/config/istio.operator.v1alpha1/#KubernetesResourcesSpec Please make a decision: do you want to keep it consistent with IstioOperator spec or not. If you decide to go with spec easily mapped to IstioOperator spec list all supported fields and make it more similar. If you go other way probably it would be better to have API more concise (remove dataplane, controlplane, deployment keywords that do not provide value)

we do not plan to expose everything in Istio CR, only what we want to support. If it makes sense some properties will be copied 1 to 1 like `numTrustedProxies`, some might be slightly moderated like `forwardClientCertDetails`, where we want to expose only some configuration. Finally we will build our dedicated configuration based on generic Istio configurations like `externalProviders`.

docs/xff-proposal.md
```diff
@@ -0,0 +1,75 @@
...
+    ingressGateway:
+      deployment:
+        hpa: 
+          maxReplicas: 5
+          minReplicas: 2
+          metrics:
+          - resource:
+              name: cpu
+              targetAverageUtilization: 80
+            type: Resource
```
videlov, line 50: this should be not be there `type: Resource`, correct ?

videlov (changes requested):

docs/xff-proposal.md
```diff
@@ -0,0 +1,75 @@
...
+Status will be implemented per `kyma-operator` requirements.
+
+## Istio CR example
+
+```
+apiVersion: operator.kyma-project.io/v1alpha1
+kind: Istio
+metadata:
+spec:
+  controlplane:
```
triffer, line 20: I think we need to adapt the fields "controlplane" and "dataplane" to use the correct camelCase form, as in the details above.

pbochynski (approved):

videlov (approved):

Issue #26, opened 2022-11-09
Library managing sidecars


strekm:
**Description**

Library should implement:
- analysing of workloads that should be part of service mesh and adding it if needed
- keeping version of sidecars up to date

ACs:
- [ ] workloads should be part of service mesh if configured
- [ ] sidecar version is up to date
- [ ] workloads are NOT labeled / annotated
- [ ] no dependency to reconciler

**Reasons**

modularization

**DoD:**
- [ ] provide unit tests

**[WIP] PR**
-  https://github.com/kyma-project/istio/pull/37

werdes72:
TODO:

- [ ] Since we don't want to label/annotate customer workloads anymore, decide what to do with existing annotations (ignore?)
- [ ] Finish pod getting logic (getting sidecar names from annotations, get image prefix and suffix from config, sidecar injection logic, cni configuration)
- [ ] Get parent objects which needs to be rolled out
- [ ] Delete pod/perform rollout

triffer:
To be discussed:
To force a rollout of pods we need to change anything in the template of the pod. In the existing implementation the annotation `kubectl.kubernetes.io/restartedAt` was added. We an AC to not annotate the workloads. How should we deal with this?

PR #37, opened 2022-12-02
Sidecar manager

werdes72:
See: #26


operator/pkg/lib/sidecars/pods/get.go
```diff
@@ -0,0 +1,151 @@
...
+
+const (
+	istioValidationContainerName = "istio-validation"
+	istioInitContainerName       = "istio-init"
+	istioSidecarName             = "istio-proxy"
+)
+
+func (r SidecarImage) matchesImageIn(container v1.Container) bool {
+	// TODO Understand why we can do a full string match
+	containsRepository := strings.Contains(container.Image, r.Repository)
```
barchw, line 24: This was used as we were using `solo.io` images at some point, but it may be possible to drop it now

operator/pkg/lib/sidecars/config.go
```diff
@@ -0,0 +1,35 @@
...
+	ImageVersion string
+
+	// Kubeclient for k8s cluster operations
+	Kubeclient kubernetes.Interface
+
+	// Debug mode
+	Debug bool
+
+	// Logger to be used
+	Log logr.Logger
```
barchw, line 28: We might want to use `zap` logger as this is the one that we are using in istio operator

operator/pkg/lib/sidecars/pods/get.go
cnvergence: Agree, we can simplify it to check the tag only.
Another idea is to check what `istioctl version` is doing to get this version.

operator/pkg/lib/sidecars/pods/filter.go
```diff
@@ -0,0 +1,139 @@
...
+	istioStatus := istioStatusStruct{}
+	err := json.Unmarshal([]byte(pod.Annotations["sidecar.istio.io/status"]), &istioStatus)
+	if err != nil {
+		return []string{}
+	}
+	return istioStatus.Containers
+}
+
+func hasSidecarContainerWithWithDifferentImage(pod v1.Pod, expectedImage SidecarImage) bool {
+	// TODO Why can't we simply assume the pod name to be istio-proxy? Because of istio-init?
```
triffer, line 49: Can someone add any information about this?

operator/pkg/lib/sidecars/pods/filter.go
```diff
@@ -0,0 +1,133 @@
+package pods
+
+import (
+	"encoding/json"
+	v1 "k8s.io/api/core/v1"
+)
+
+const AnnotationResetWarningKey = "istio.reconciler.kyma-project.io/proxy-reset-warning"
```
triffer, line 8: As far as I remember we don't want to annotate customer workload any longer. That means that the restart won't set it and therefore we don't need to use it as a filter.

operator/pkg/lib/sidecars/proxy.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"context"
+	"github.com/go-logr/logr"
+	"github.com/kyma-project/istio/operator/pkg/lib/sidecars/pods"
+	"github.com/kyma-project/istio/operator/pkg/lib/sidecars/restart"
+	v1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+//
```
triffer, line 12: I think we can remove this as this is outdated.

operator/pkg/lib/sidecars/pods/get.go
```diff
@@ -0,0 +1,151 @@
...
+			outputPodsList.Items = append(outputPodsList.Items, *pod.DeepCopy())
+		}
+	}
+
+	return outputPodsList, nil
+}
+
+func GetPodsForCNIChange(ctx context.Context, c client.Client, isCNIEnabled bool) (outputPodsList v1.PodList, err error) {
+	podList, err := getAllRunningPods(ctx, c)
+	// TODO: add logs
```
triffer, line 92: There are two TODOs about adding logs. Maybe we can check what logs what logs are implemented in the old code and evaluate if they are useful.

operator/pkg/lib/sidecars/pods/get_test.go
```diff
@@ -0,0 +1,684 @@
...
+				NewSidecarPodBuilder().SetName("application1").SetNamespace("kube-system").Build(),
+				NewSidecarPodBuilder().SetName("application2").SetNamespace("kube-public").Build(),
+				NewSidecarPodBuilder().SetName("application3").SetNamespace("istio-system").Build(),
+				NewSidecarPodBuilder().SetName("application4").SetNamespace("enabled").Build(),
+				enabledNamespace,
+			),
+			expectedImage: pods.SidecarImage{Repository: "istio/proxyv2", Tag: "1.10.0"},
+			isCNIEnabled:  true,
+			wantEmpty:     false,
+			wantLen:       1,
```
triffer, line 69: I think it would be better to not only rely on the size of the returned result, but also verify that the correct result is returned, e.g. by asserting the namespace.

operator/pkg/lib/sidecars/pods/get_test.go
```diff
@@ -0,0 +1,684 @@
...
+			c: createClientSet(t,
+				NewSidecarPodBuilder().SetName("application1").SetNamespace("enabled").Build(),
+				NewSidecarPodBuilder().SetName("application2").SetNamespace("enabled").
+					SetPodStatusPhase("Terminating").Build(),
+				enabledNamespace,
+			),
+			expectedImage: pods.SidecarImage{Repository: "istio/proxyv2", Tag: "1.10.0"},
+			isCNIEnabled:  true,
+			wantEmpty:     false,
+			wantLen:       1,
```
triffer, line 94: I think it would be better to not only rely on the size of the returned result, but also verify that the correct result is returned, e.g. by asserting the pod status phase.

operator/pkg/lib/sidecars/restart/replica_set_action.go
```diff
@@ -0,0 +1,48 @@
...
+	err := c.Get(ctx, replicaSetKey, replicaSet)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			return newOwnerNotFoundAction(pod), nil
+		}
+		return restartAction{}, err
+	}
+
+	if rsOwnedBy, exists := getReplicaSetOwner(replicaSet); !exists {
+		return newDeleteAction(actionObjectFromPod(pod)), nil
```
triffer, line 31: We could add a comment why in this case the pod will be deleted, to make it easier to understand without thinking about details.

operator/pkg/lib/sidecars/pods/filter.go
cnvergence: We can safely assume that container name will be `istio-proxy`

operator/pkg/lib/sidecars/restart/restart_test.go
```diff
@@ -0,0 +1,255 @@
...
+		require.NotEmpty(t, warnings)
+
+		pods := v1.PodList{}
+		err = c.List(context.TODO(), &pods)
+
+		require.NoError(t, err)
+		require.NotEmpty(t, pods.Items)
+	})
+
+	t.Run("should not delete pod when it is owned by a ReplicaSet that is found", func(t *testing.T) {
```
triffer, line 210: I think to align with the other tests this can be changed to `should rollout restart ReplicaSet if the pod is owned by one`. This also means that the Pod is not deleted.

operator/pkg/lib/sidecars/restart/rollout_action.go
```diff
@@ -0,0 +1,63 @@
...
+	case "DaemonSet":
+		obj = &appsv1.DaemonSet{}
+	case "Deployment":
+		obj = &appsv1.Deployment{}
+	case "ReplicaSet":
+		obj = &appsv1.ReplicaSet{}
+	case "StatefulSet":
+		obj = &appsv1.StatefulSet{}
+	default:
+		return nil, fmt.Errorf("kind %s not found", object.Kind)
```
triffer, line 38: Maybe "not found" can be misleading. I think "not supported" gives a better idea that an implementation might be missing or that a wrong kind reached this code.

operator/pkg/lib/sidecars/restart/rollout_action.go
```diff
@@ -0,0 +1,63 @@
...
+
+	obj.SetAnnotations(annotations)
+
+	err = k8sclient.Update(ctx, obj)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return nil, nil
```
triffer, line 62: In this case I'd rather return an empty array instead of nil. Because if error is not nil, I think you should expect to have a valid value for the first result.

barchw: In go len(nil) == 0, so this is OK imho

operator/pkg/lib/sidecars/pods/filter.go
werdes72: Removed

triffer: Discussion outcome: We can remove it.

operator/pkg/lib/sidecars/pods/get_test.go
triffer: Discussion outcome: We should add the additional assertion.

operator/pkg/lib/sidecars/pods/get_test.go
```diff
@@ -0,0 +1,684 @@
...
+			c: createClientSet(t,
+				NewSidecarPodBuilder().SetName("application1").SetNamespace("enabled").Build(),
+				NewSidecarPodBuilder().SetName("application2").SetNamespace("enabled").
+					SetInitContainer("istio-validation").Build(),
+				enabledNamespace,
+			),
+			expectedImage: pods.SidecarImage{Repository: "istio/proxyv2", Tag: "1.10.0"},
+			isCNIEnabled:  true,
+			wantEmpty:     false,
+			wantLen:       1,
```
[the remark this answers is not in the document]
triffer: Discussion outcome: We should add the additional assertion.

operator/pkg/lib/sidecars/restart/replica_set_action.go
triffer: Discussion outcome: We should add a simple comment.

operator/pkg/lib/sidecars/restart/restart_test.go
triffer: Discussion outcome: Add the `should rollout restart ReplicaSet if the pod is owned by one` as a new separate test and keep the `should not delete pod when it is owned by a ReplicaSet that is found` test.

operator/pkg/lib/sidecars/restart/rollout_action.go
```diff
@@ -0,0 +1,63 @@
...
+	annotations := obj.GetAnnotations()
+	if len(annotations) == 0 {
+		annotations = map[string]string{}
+	}
+
+	annotations[annotationName] = time.Now().Format(time.RFC3339)
+
+	obj.SetAnnotations(annotations)
+
+	err = k8sclient.Update(ctx, obj)
```
cnvergence, line 56: Maybe let's use the Patch method here with the StrategicMergePatch type here, to just apply the annotations patch instead of the whole object.

operator/pkg/lib/sidecars/proxy.go
```diff
@@ -0,0 +1,49 @@
...
+		return nil, err
+	}
+
+	var podListToRestart v1.PodList
+	podListToRestart.Items = []v1.Pod{}
+	differentImagePodList.DeepCopyInto(&podListToRestart)
+	podListToRestart.Items = append(podListToRestart.Items, noSidecarPodList.DeepCopy().Items...)
+	podListToRestart.Items = append(podListToRestart.Items, cniPodList.DeepCopy().Items...)
+
+	for _, pod := range differentImagePodList.Items {
```
barchw, line 35: Seems redundant now

operator/pkg/lib/sidecars/proxy.go
```diff
@@ -0,0 +1,49 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func ProxyReset(ctx context.Context, c client.Client, expectedImage pods.SidecarImage, namespaceEnabledByDefault, cniEnabled bool, logger *logr.Logger) ([]restart.RestartWarning, error) {
+	differentImagePodList, err := pods.GetPodsWithDifferentSidecarImage(ctx, c, expectedImage)
+	if err != nil {
+		return nil, err
+	}
+
+	noSidecarPodList, err := pods.GetPodsWithoutSidecar(ctx, c, namespaceEnabledByDefault)
```
barchw, line 19: Looks to me as though we are missing a negative case (pods that have sidecar right now but shouldn't)

operator/pkg/lib/sidecars/restart/restart_test.go
```diff
@@ -0,0 +1,304 @@
...
+
+		// then
+		require.NoError(t, err)
+		require.Empty(t, warnings)
+
+		obj := appsv1.Deployment{}
+		err = c.Get(context.TODO(), types.NamespacedName{Namespace: "test-ns", Name: "owner"}, &obj)
+		require.NoError(t, err)
+
+		require.NotEmpty(t, "owner", obj.Annotations[annotationName])
```
triffer, line 108: I think the asserts of the annotation don't work as expected. As the function signature is `NotEmpty(t TestingT, object interface{}, msgAndArgs ...interface{})`. That means in this case `owner` is the tested object and `obj.Annotations[annotationName]` is passed as `msgAndArgs`.
We might also use the `ResourceVersion` of the object to assert if it was changed.

operator/pkg/lib/sidecars/restart/rollout_action.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"github.com/kyma-project/istio/operator/pkg/lib/sidecars/retry"
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const rolloutTimeoutMessage = "pod could not be rolled out by resource owner's controller."
+
+const annotationName = "kubectl.kubernetes.io/restartedAt"
```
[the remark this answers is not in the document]
barchw:
```suggestion
const annotationName = "istio-operator.kyma-project.io/restartedAt"
```

operator/pkg/lib/sidecars/proxy.go
cnvergence: yes, so far we have supported only pods that should had a sidecar but they didn't

operator/pkg/lib/sidecars/restart/restart_test.go
barchw: https://github.com/kyma-project/istio/pull/37/commits/def272fdf7cee61fd9bb8ad2ac6616ec3acb9228

operator/pkg/lib/sidecars/pods/get_test.go
werdes72: Added asserting pod name.

operator/pkg/lib/sidecars/proxy.go
triffer: If it was not relevant in the past, do we have evidence that this is a valid case to support in the future?

barchw: I was sure that we are supporting this in reconciler :O
We should discuss this functionality then

operator/pkg/lib/sidecars/pods/get_test.go
werdes72: Is checking for podName enough or phase/namespace should also be asserted?

operator/pkg/lib/sidecars/proxy.go
cnvergence: It was mainly implemented to fix the issue of missing sidecar injection(mutating webhook), meaning that normally if you change label/annotation to disable, it will get restarted automatically.
With the default strategy of sidecar injection being opt-in instead of opt-out, it doesn't feel crucial to support this the same way.

But I see one case that we can extend, when you have a deployment with sidecar injected and you change its namespace label to `istio-injection=disabled`.

triffer: I think we should add a new issue for this and discuss it in the refinement and then decide if we want to implement it or not.

operator/pkg/lib/sidecars/restart/restart.go
```diff
@@ -0,0 +1,53 @@
...
+	for _, pod := range podList.Items {
+		action, err := restartActionFactory(ctx, c, pod)
+		if err != nil {
+			return nil, err
+		}
+
+		// We want to avoid performing the same action multiple times for a parent if it contains multiple pods that need to be restarted.
+		if _, exists := processedActionObjects[action.object.getKey()]; !exists {
+			currentWarnings, err := action.run(ctx, c, action.object, logger)
+			if err != nil {
```
triffer, line 43: With the current implementation we fail fast the restarting of all pods once there is an error for any pod.
Do we want to be more resilient? I think we also didn't do it in the implementation in reconciler.
We can log the error and create a restart warning for the pod here or we could do it in each run function.

operator/pkg/lib/sidecars/proxy.go
barchw: I will adapt the component test for now and leave a note in case we want to extend that

barchw: https://github.com/kyma-project/istio/pull/37/commits/10d33020c88abfdbc2ac347f06250f0d8f5574da

operator/pkg/lib/sidecars/restart/restart.go
```diff
@@ -27,21 +27,21 @@ func newRestartWarning(o actionObject, message string) RestartWarning {
...
 func Restart(ctx context.Context, c client.Client, podList v1.PodList, logger *logr.Logger) ([]RestartWarning, error) {
-
 	warnings := make([]RestartWarning, 0)
 	processedActionObjects := make(map[string]bool)
 
 	for _, pod := range podList.Items {
 		action, err := restartActionFactory(ctx, c, pod)
 		if err != nil {
-			return nil, err
+			logger.Error(err, "creating an action for a pod failed")
```
barchw, line 36: Can't we make it a warning? This would make more sense imho

operator/pkg/lib/sidecars/restart/restart.go
werdes72: Added logging the error without breaking the loop.

operator/pkg/lib/sidecars/proxy.go
barchw: As discussed in refinement removed support for missing sidecars in https://github.com/kyma-project/istio/pull/37/commits/50d9428e802a2b8778891ac96c68aea917c5de6b

barchw (approved):
/bark

barchw:
@werdes72 @cnvergence Please approve and uphold if you don't need anything more

Issue #4, opened 2022-06-14
Basic version of kyma istio operator reacting on CR creation


strekm:
**Description**

Implement first version of kyma istio operator which reacts on creation of CR. Istioctl should NOT be used to install Istio. It is possible to configure `numTrustedProxies`. Applying profiles is out of scope of this task. Base Istio configuration from kyma should used for installation.

AC:
- [x] kyma istio operator installs istio based on conf in Istio CR
- [x] documentation created
- [x] manual test case created

**Reasons**

Modularisation of kyma components

**Attachments**
https://github.com/kyma-project/istio/blob/main/charts/istio-operator/crds/apiextensions.k8s.io_v1_customresourcedefinition_istios.operator.kyma-project.io.yaml
https://github.com/kyma-project/kyma/blob/main/resources/istio/templates/istio-operator.yaml
depends on: https://github.com/orgs/kyma-project/projects/27/views/1

strekm:
for sake of simplicity we assume that IstioOperatorFile CR is hardcoded in Istio operator. Version of Istio is also hardcoded.

PR #54, opened 2023-01-11
Initial version of Kyma Istio Operator

cnvergence:
**Description**

- ...
- ...
- ...

**Related issue(s)**

See also:
- https://github.com/kyma-project/istio/issues/4


operator/Dockerfile
```diff
@@ -29,6 +31,8 @@ FROM gcr.io/distroless/static:nonroot
 WORKDIR /
 COPY istio-chart/ istio-chart/
 COPY --from=builder /workspace/manager .
+COPY --from=builder /workspace/manifests/ .
```
barchw, line 34: Maybe we should make this directory identical as for running locally? For now `make run` doesn't work
```suggestion
COPY --from=builder /workspace/manifests/ manifests
```

operator/pkg/manifest/manifest.go
```diff
@@ -0,0 +1,54 @@
...
+	"sigs.k8s.io/yaml"
+)
+
+var (
+	defaultIstioOperatorFile = "default-istio-operator-k3d.yaml"
+)
+
+// Merge applies configuration from IstioCR to the default Kyma Istio Operator file, saves it to a file and returns its path.
+func Merge(istioCR *operatorv1alpha1.Istio) (string, error) {
+	manifest, err := os.ReadFile(defaultIstioOperatorFile)
```
barchw, line 18: Adding manifestDir const might also be an idea
```suggestion
	manifest, err := os.ReadFile("manifests", defaultIstioOperatorFile)
```

operator/pkg/manifest/manifest.go
```diff
@@ -0,0 +1,54 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	mergedManifest, err := applyIstioCR(istioCR, manifest)
+	if err != nil {
+		return "", err
+	}
+	mergedManifestFilePath := fmt.Sprintf("/tmp/%s", defaultIstioOperatorFile)
```
barchw, line 27:
```suggestion
	mergedManifestFilePath := path.Join("/tmp", defaultIstioOperatorFile)
```

operator/pkg/manifest/manifest.go
```diff
@@ -0,0 +1,54 @@
...
+	"fmt"
+	"os"
+
+	operatorv1alpha1 "github.com/kyma-project/istio/operator/api/v1alpha1"
+	istioOperator "istio.io/istio/operator/pkg/apis/istio/v1alpha1"
+	"sigs.k8s.io/yaml"
+)
+
+var (
+	defaultIstioOperatorFile = "default-istio-operator-k3d.yaml"
```
barchw, line 13: We could probably use https://github.com/kyma-project/cli/blob/16c49248da0432e7786757bac132d9741ee2a29e/internal/clusterinfo/clusterinfo.go#L43 for checking on which cluster we are
```suggestion
	defaultIstioOperatorFile = "default-istio-operator-k3d.yaml"
```

barchw (changes requested):

operator/controllers/istio_controller.go
```diff
@@ -41,11 +45,34 @@ func TemplateRateLimiter(failureBaseDelay time.Duration, failureMaxDelay time.Du
...
+	istioCR := operatorv1alpha1.Istio{}
+	if err := r.Client.Get(ctx, req.NamespacedName, &istioCR); err != nil {
+		logger.Error(err, "Error during fetching Istio CR")
+	}
+
+	if err := install.ReconcileIstio(&istioCR); err != nil {
+		logger.Error(err, "Error occurred during reconciliation of Istio Operator")
+	}
+
+	return ctrl.Result{}, nil
```
barchw, line 69: https://github.tools.sap/xf-goat/adr/blob/main/doc/adr/0003-istio-operator-design.md#reconciliation-interval
```suggestion
	return ctrl.Result{RequeueAfter: time.Minute*5}, nil
```

triffer:
The implementation doesn't follow the [technical design](https://github.com/kyma-project/istio/blob/main/docs/technical-design.md#components) we created. There are components like [IstioInstallationReconciliation](https://github.com/kyma-project/istio/blob/main/docs/technical-design.md#istioinstallationreconciliation) and [IstioClient](https://github.com/kyma-project/istio/blob/main/docs/technical-design.md#istioclient) that should be part of this implementation.

operator/manifests/default-istio-operator-cluster.yaml
```diff
@@ -0,0 +1,319 @@
+apiVersion: install.istio.io/v1alpha1
+kind: IstioOperator
+metadata:
+  name: installed-state-default-operator
```
barchw, line 4: Istio installation adds `installed-state-` prefix
```suggestion
  name: default-operator
```

operator/pkg/manifest/manifest.go
cnvergence: I have just added a second manifest for the future, so we won't need to look at what we are missing.
This is currently out of the scope of this task.
That's a good idea, we will need to extend our configuration, so I think we should check it and apply different CNI config if needed.

operator/Dockerfile
cnvergence: True, I will fix this

operator/controllers/istio_controller.go
cnvergence: Good point!

barchw (approved):
/hold Waiting for @triffer

Issue #81, opened 2023-02-15
Uninstallation of Istio


strekm:
**Description**

Add support for Istio deletion. When user deletes Istio CR deletion of Istio should be triggered. 

- [x] istio uninstalls on CR deletion
- [x] no leftovers (e.g. CRD, namespace)
- [x] it is possible to install Istio after uninstall

**Reasons**

modularisation

**DoD:**
- [x] provide documentation
- [x] provide unit tests
- [x] test on SKR / gardener
~~- [ ] verify resource limits~~
~~- [ ] followup issue~~
~~- [ ] publish release notes and _What's New_ updates for Kyma customers~~

**PR:**
- [x] https://github.com/kyma-project/istio/pull/90

**Attachments**
part of: https://github.com/kyma-project/istio/issues/38

triffer:
There are a few things to keep in mind as this behavior is currently expected of us.

- When there is a workload in namespace with Istio enabled, this workload won't get the proxy after Istio is installed. This is the same behaviour as when using istioctl.

- When Istio is installed and there is a workload with a proxy then this proxy will stay when Istio is unistalled. This is the same behaviour as when using istioctl.

- Proxy reset will also reset manager on update when the manager is deployed in kyma-system namespace, do we expect this to be an issue. Do we need to exclude the manager from sidecar injection?

PR #90, opened 2023-02-22
Uninstallation of Istio

triffer:
**Description**

- Add handling for uninstalling Istio when Istio CR is deleted

**Related issue(s)**

- https://github.com/kyma-project/istio/issues/81


operator/internal/reconciliations/istio/installation.go
```diff
@@ -1,30 +1,99 @@
...
+
+	// We need to evaluate what changed since last reconciliation, to make sure we run Istio reconciliation only if it's necessary
+	istioCRChanges, err := EvaluateIstioCRChanges(istioCR, istioTag)
 	if err != nil {
-		return err
+		ctrl.Log.Error(err, "Error evaluating IstioCR changes")
+		return istioCR, err
+	}
+
+	if !istioCRChanges.MustBeReconciled() {
```
cnvergence, line 37: do we need this check at all?

triffer: That's a good point, I think we can remove it.

Issue #82, opened 2023-02-15
Module build pipeline


strekm:
**Description**

Implement prow pipeline in https://github.com/kyma-project/test-infra repository based on keda manager that builds Istio module. Pipeline depends on build pipeline. It should be pre / post main mandatory pipeline. Alpha artefacts (module image, module template) should be pushed  dev registry. There should be no PR to kyma repository pushing module template.

ACs:
- [x] module image is available in registry
- [x] module templates are stored in dev bucket --> Will be handled in https://github.com/kyma-project/istio/issues/108
- [x] module can be installed by lifecycle manager

**Reasons**

modularisation

**DoD:**
~~- [ ] provide documentation~~
~~- [ ] provide unit tests~~
- [x] test on SKR / gardener
~~- [ ] verify resource limits~~
~~- [ ] followup issue~~
~~- [ ] publish release notes and _What's New_ updates for Kyma customers~~

**Attachments**
part of: https://github.com/kyma-project/istio/issues/38
based on POC results

**PRs**
- https://github.com/kyma-project/test-infra/pull/7090
- https://github.com/kyma-project/istio/pull/96

PR #96, opened 2023-02-28
Flatten operator directories and align with Keda module frontrunner

barchw:
**Description**

- Align our module with Keda frontrunner - move operator directory to repository root
- Add makefiles for module building

**Related issue(s)**

- https://github.com/kyma-project/istio/issues/82

**Test infra PR**
- https://github.com/kyma-project/test-infra/pull/7090


.env
```diff
@@ -0,0 +1,3 @@
+ifndef MODULE_VERSION
```
triffer, line 1: Isn't it confusing that we have defaults defined in the Makefile itself and the .env file?
What's the benefit for us?

triffer (approved):
Good job, thanks 👍

.env
barchw: You're right, it might be a little confusing, we can adapt that if we want to

Issue #106, opened 2023-03-06
Perf tests should not have dependency to api-gateway nor function


strekm:
**Description**

Istio being core component should NOT have dependency to any other kyma components (api-gateway, functions). It should not be required to install any other components to execute performance test (or any other tests). It should be possible to execute perf tests (or any other int tests) on upcoming Istio version to see results before releasing Istio module. authn/authz should not be tested in this perf tests (that might be tested in api-gateway in future). Remove dependency to api-gateway and functions from perf test. Remove any unused resources. Remove dependency to oauth2client from ORY Hydra.

ACs:
- [x] tests can be executed on a cluster with Istio (no dependency to any other kyma components)
- [x] there are no unused resources in tests
- [x] clear documentation on what is tested
- [x] httpbin exposed via VS only

**PRs**
- https://github.com/kyma-project/istio/pull/116
- https://github.com/kyma-project/api-gateway/pull/256

strekm:
existing implementation can be moved to api-gateway

PR #116, opened 2023-03-14
Use only istio in performance tests

barchw:
**Description**

- Remove dependencies to Ory, API-Gateway
- Expose workload with VS
- Adapt Grafana dashboard to export Istio related metrics
- Adapt `common.js` script for allow strategy, removing OAuth2 token fetching
- Add deployment of Prometheus to Chart

**Related issue(s)**

- https://github.com/kyma-project/istio/issues/106


performance_tests/Makefile
```diff
@@ -0,0 +1,36 @@
+OS_ARCH ?= $(shell uname | awk '{print tolower($0)}')
+KYMA_DOMAIN ?= $(k config view -o json | jq '.clusters[0].cluster.server' | sed -e "s/https:\/\/api.//" -e 's/"//g')
```
werdes72, line 2: I would use `kubectl` here as not everyone has this alias.

Issue #128, opened 2023-04-13
Blocking deletion strategy


strekm:
**Description**

Based on [decision](https://github.com/kyma-project/community/issues/765) made implement blocking deletion strategy for istio module. Istio should not be deleted if there are Istio related CRs present on a cluster. 

If there is any Istio related resource on a cluster Istio CR should be in error state, finaliser should not be released. On error reconciliation will be retried after 10 mins. It is expected that an user will delete all istio resources or modules depending on istio resources. Once all Istio resources are gone deletion can be continued, no retry from used side is needed. 

List of resources to check should be based on Istio CRDs. Resources created by `istio-resources` chart (in the future self created resources) should be excluded from the check. 

ACs:
- [x] istio is deleted if there are not istio CRs on a cluster
- [x] Istio CR is in error state and have finaliser if there are Istio CRs

**Reasons**

modularisation

**PR**
- https://github.com/kyma-project/istio/pull/129

**DoD:**
- [x] provide documentation
~~- [ ] release notes and What's New updates for Kyma customers~~
- [x] provide unit tests
- [x] provide scenario for integration tests
- [x] test on production-like environment
~~- [ ] verify resource limits~~
- [x] [followup issue created](https://github.com/kyma-project/istio/issues/130)

**Attachments**
https://github.com/kyma-project/community/issues/765
part of: https://github.com/kyma-project/istio/issues/38

PR #129, opened 2023-04-14
Blocking uninstall of Istio if customer resources are present

barchw:
**Description**

- Block Istio uninstall if customer resources are present

**Related issue(s)**

- https://github.com/kyma-project/istio/issues/128


internal/reconciliations/istio/installation.go
```diff
@@ -100,7 +102,23 @@ func (i *Installation) Reconcile(ctx context.Context, client client.Client, isti
...
+		}
+
+		clientResources, err := istioResourceFinder.FindUserCreatedIstioResources()
+		if err != nil {
+			return istioCR, err
+		}
+		if len(clientResources) > 0 {
+			errorTemplate := "Could not delete Istio module instance since there are customer created resources present:"
+			for _, resource := range clientResources {
+				errorTemplate += fmt.Sprintf("\n%s:%s/%s", resource.GVK.Kind, resource.Namespace, resource.Name)
```
videlov, line 117: listing all recourses may create a big list on big clusters. although it will be helpful, I would suggest to limit it just kind of resources (and eventually count) rather to list them all

internal/resources/resources.go
```diff
@@ -0,0 +1,88 @@
...
+		configuration: finder,
+	}, nil
+}
+
+func (i *IstioResourcesFinder) FindUserCreatedIstioResources() ([]Resource, error) {
+	var userResources []Resource
+	for _, resource := range i.configuration.Resources {
+		var u unstructured.UnstructuredList
+		u.SetGroupVersionKind(resource.GroupVersionKind)
+		err := i.client.List(i.ctx, &u)
```
videlov, line 62: this might be not quite the best in large clusters, but since it is an uninstall do not happens often and perhaps not so much user resources at this step, and I don't have better solution, let have it like that

internal/resources/resources_test.go
```diff
@@ -0,0 +1,138 @@
...
+						},
+					},
+				},
+			},
+		},
+		))
+	})
+})
+
+var _ = DescribeTable("contains", func(a []ResourceMeta, b Resource, should bool) {
```
videlov, line 93: this because of code-coverage gate right? 🥇

manifests/controlled_resources_list.yaml
```diff
@@ -0,0 +1,77 @@
+resources:
```
videlov, line 1: Do we need to add all Istio CRs on an empty installtion of Istio on top of Kyma ? I see more CRs on my instance. We must test this with an integration test, otherwise we risk uninstallation to fail quite recently because we base it on a static list of resources. Unit tested is not enough.

internal/resources/resources_test.go
```diff
@@ -0,0 +1,138 @@
...
+			ctx:           ctx,
+			logger:        logger,
+			client:        client,
+			configuration: configuration,
+		}
+		got, err := i.FindUserCreatedIstioResources()
+		Expect(err != nil).To(Equal(wantErr))
+		Expect(got).To(BeEquivalentTo(want))
+	},
+		Entry("Should get nothing if there are only default istio resources present", context.TODO(),
```
videlov, line 34: should we add also test when we find a customer created resource?

videlov (changes requested):

manifests/controlled_resources_list.yaml
```diff
@@ -0,0 +1,77 @@
...
+        namespace: "istio-system"
+      - name: "stats-filter-1.17"
+        namespace: "istio-system"
+      - name: "stats-filter-1.18"
+        namespace: "istio-system"
+      - name: "tcp-stats-filter-1.16"
+        namespace: "istio-system"
+      - name: "tcp-stats-filter-1.17"
+        namespace: "istio-system"
+      - name: "tcp-stats-filter-1.18"
```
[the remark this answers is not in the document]
barchw: Yeah, I was also thinking about this one, but since we will be providing new version of operator for new Istio version it's not a technical problem. We might make it more flexible (like using regex and looking for `tcp-stats-filter-\d.\d\d"), but this might change in Istio resources so I don't have any strong opinion about this.
If you feel like it would be better, I will switch the match to use regex

internal/resources/resources_test.go
barchw: 🥇

manifests/controlled_resources_list.yaml
barchw: Discussed this one, for now let's leave only Istio Operator created resources, we can extend it later

internal/resources/resources_test.go
barchw: Yes

manifests/controlled_resources_list.yaml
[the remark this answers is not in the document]
barchw: Changed the match to regex

videlov (dismissed):

videlov (approved):

Issue #143, opened 2023-05-12
Resource and HPA settings support


strekm:
**Description**
Enhance Istio CRD with components spec as designed in [Istio CRD spec proposal](https://github.com/kyma-project/istio/blob/main/docs/xff-proposal.md). It should be possible to configure resources, affinity and hpa. This configuration should be possible for cni, pilot, ingressGateway and  proxy.

At this moment default Istio configuration should be enhanced with respective values from kyma profile configuration. Please be aware of https://github.com/kyma-project/istio/issues/110.

ACs:
- [x] it is possible to configure resources, affinity and hpa for cni, pilot, ingressGateway and proxy
- [x] validation implemented
~~- [ ] default values updated~~ - The default values are coming from production configuration and overridden if the config is present

**Reasons**

**DoD:**
- [x] provide documentation
~- [ ] release notes and What's New updates for Kyma customers~
- [x] provide unit tests
- [x] provide integration tests
- [x] test on production-like environment
- [x] verify resource limits
- [ ] [followup issue](https://github.com/kyma-project/istio/issues/150)

**Attachments**
part of: https://github.com/kyma-project/istio/issues/38

**PR**
- [x] https://github.com/kyma-project/istio/pull/145

PR #145, opened 2023-05-17
Add configuration for resources to Istio Custom Resource

barchw:
**Description**

- Add support for configuring Istio resources in CR

**Related issue(s)**

- https://github.com/kyma-project/istio/issues/143


internal/reconciliations/istio/configuration.go
```diff
@@ -78,6 +83,108 @@ func EvaluateIstioCRChanges(istioCR operatorv1alpha1.Istio, istioTag string) (tr
 	return trigger, nil
 }
 
+func checkComponentsConfigChange(components *operatorv1alpha1.Components, components2 *operatorv1alpha1.Components) IstioCRChange {
```
triffer, line 86: Why does this "check" function return a IstioCRChange and checkK8SConfigChange a boolean? I think it would be easier to use booleans in both cases to keep IstioCRChange out of this level of the functions.

internal/reconciliations/istio/configuration.go
```diff
@@ -78,6 +83,108 @@ func EvaluateIstioCRChanges(istioCR operatorv1alpha1.Istio, istioTag string) (tr
...
+			if checkK8SConfigChange(ingressGateway.K8s, components2.IngressGateways[i].K8s) {
+				return ConfigurationUpdate
+			}
+		}
+	}
+
+	return NoChange
+}
+
+func checkK8SConfigChange(config operatorv1alpha1.KubernetesResourcesConfig, config2 operatorv1alpha1.KubernetesResourcesConfig) bool {
```
triffer, line 108: I think this helps understanding what return values we can expect.
```suggestion
func hasK8SConfigChange(config operatorv1alpha1.KubernetesResourcesConfig, config2 operatorv1alpha1.KubernetesResourcesConfig) bool {
```

internal/reconciliations/istio/configuration.go
```diff
@@ -60,14 +60,19 @@ func EvaluateIstioCRChanges(istioCR operatorv1alpha1.Istio, istioTag string) (tr
 		trigger = trigger | VersionUpdate
 	}
 
-	lastAppliedNotNil := lastAppliedConfig.Config.NumTrustedProxies != nil
-	newNotNil := istioCR.Spec.Config.NumTrustedProxies != nil
+	if nilChange(lastAppliedConfig.Components, istioCR.Spec.Components) {
+		return trigger | ConfigurationUpdate, nil
```
triffer, line 64: I think we should simplify the IstioCRChange a lot, since we only care about two information -> `install with given configuration` or `delete`. All this extra states that are hidden in the int are not relevant right now and therefore we should remove it.

tests/integration/istio_instalation_scenario.go
triffer, line 1: typo in file name, `instalation` -> `installation`

Makefile
```diff
@@ -108,7 +108,7 @@ vet: ## Run go vet against code.
 
 .PHONY: test
 test: manifests generate fmt vet envtest ## Run tests.
-	KUBEBUILDER_CONTROLPLANE_START_TIMEOUT=2m KUBEBUILDER_CONTROLPLANE_STOP_TIMEOUT=2m KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" go test $(shell go list ./... | grep -v /tests/integration) -coverprofile cover.out
+	KUBEBUILDER_CONTROLPLANE_START_TIMEOUT=2m KUBEBUILDER_CONTROLPLANE_STOP_TIMEOUT=2m KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" go test $(shell go list ./... | grep -v controllers | grep -v /tests/integration) -coverprofile cover.out
```
triffer, line 111: I think we might also want to consider excluding `pkg/lib/sidecars/test`, `tests` and since we can't easily ignore generated code also `api`. On the other hand, with such a configuration it is also very questionable whether we should keep the code coverage check.

api/v1alpha1/merge_test.go
```diff
@@ -174,6 +175,229 @@ var _ = Describe("Merge", func() {
 			GetStructValue().Fields["gatewayTopology"].GetStructValue().Fields["numTrustedProxies"].GetNumberValue()
 		Expect(numTrustedProxies).To(Equal(float64(5)))
 	})
+
```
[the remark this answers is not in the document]
barchw: 1. Sure, we can have such test
2. I would say that this is covered by integration test scenario, as we cannot fully rely on the IOP in terms of what is going to be applied

Makefile
barchw: I think we need to discuss this code coverage check

api/v1alpha1/merge_test.go
```diff
@@ -325,7 +327,7 @@ var _ = Describe("Merge", func() {
 	})
 
 	Context("Strategy", func() {
-		It("Should update RollingUpdate when it is present in Istio CR", func() {
+		It("Should update &RollingUpdate when it is present in Istio CR", func() {
```
triffer, line 330:
```suggestion
		It("Should update RollingUpdate when it is present in Istio CR", func() {
```

api/v1alpha1/merge_test.go
```diff
@@ -398,6 +400,244 @@ var _ = Describe("Merge", func() {
...
+				Expect(iopCpuRequests).To(Equal(cpuRequests))
+
+				iopMemoryRequests := out.Spec.Components.Cni.K8S.Resources.Requests["memory"]
+				Expect(iopMemoryRequests).To(Equal(memoryRequests))
+			})
+		})
+	})
+
+	Context("Proxy", func() {
+		It("Should update Proxy resources configuration if they are present in Istio CR", func() {
```
triffer, line 574: This test has exactly the same name as the following Proxy test.
Do we need this test, since the other test also sets the requests?

internal/reconciliations/istio/configuration.go
```diff
@@ -60,131 +60,17 @@ func EvaluateIstioCRChanges(istioCR operatorv1alpha1.Istio, istioTag string) (tr
 		trigger = trigger | VersionUpdate
 	}
 
-	if nilChange(lastAppliedConfig.Components, istioCR.Spec.Components) {
+	if !cmp.Equal(lastAppliedConfig.Components, istioCR.Spec.Components) {
```
triffer, line 63: I was not aware of the existence of this lib. I like this so much better, good job!

api/v1alpha1/merge_test.go
barchw: We don't :)

Makefile
barchw: Code coverage check will be optional

triffer (dismissed):

Issue #149, opened 2023-05-23
Enable warden validation for `istio-system` namespace


strekm:
**Description**

`istio-system` namespace is managed namespace by istio module (atm by istio reconciler). Kyma users should not be able to deploy to `istio-system` namespace any workloads. Enable warden on `istio-system` namespace by labeling it with warden specific label. Consider adding warning label that this namespace is managed by istio module and any changes might be reverted. As a stretch add warden label in istio reconciler.

ACs:
- [x] `istio-system` namespace is labeled with warden label
- [x] `istio-system` namespace is annotated with info that it's managed by istio module
- [x] it is not possible to deploy to `istio-system` not signed image

**Reasons**

prevent harmful modification in `istio-system`

**DoD:**
- [x] provide documentation
~~- [ ] release notes and What's New updates for Kyma customers~~
- [x] provide unit tests
~~- [ ] provide integration tests~~
- [x] test on production-like environment
~~- [ ] verify resource limits~~
~~- [ ] followup issue~~

**Attachments**
https://github.com/kyma-project/warden/blob/8a4d2467496e3f0aca029d8a5752c53770028124/pkg/labels.go#L4

PR #162, opened 2023-06-05
Label Istio namespace after installation

werdes72:
Part of #149 

Label `istio-system` namespace with warden config and also with info about being managed by Kyma.

### DIFF
--- a/internal/reconciliations/istio/installation.go
+++ b/internal/reconciliations/istio/installation.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"github.com/kyma-project/istio/operator/internal/resources"
+	"github.com/kyma-project/istio/operator/pkg/lib/label"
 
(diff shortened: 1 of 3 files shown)


tests/integration/steps/resource.go
```diff
@@ -98,6 +98,25 @@ func NamespaceIsPresent(ctx context.Context, name, shouldBePresent string) error
...
 
+func NamespaceHasLabelAndAnnotation(ctx context.Context, name, label, annotation string) error {
+	k8sClient, err := testcontext.GetK8sClientFromContext(ctx)
+	if err != nil {
+		return err
+	}
+
+	var ns corev1.Namespace
+	return retry.Do(func() error {
+		err := k8sClient.Get(context.TODO(), types.NamespacedName{Name: name}, &ns)
```
triffer, line 109: I think we should check err, because I think we can run into a panic by calling ns.Labels[label] when the ns is not found.

triffer (approved):

PR #163, opened 2023-06-05
Add script for running istio integration tests on gardener

barchw:
**Description**

- Prepare make command and shell script for provisioning Gardener cluster and running Istio tests
- Script is ready for running on different cluster providers so it can be reused for autoscaler compliancy check later

**Related issue(s)**

- https://github.com/kyma-project/test-infra/issues/7564

### DIFF
--- a/hack/ci/jobguard.sh
+++ b/hack/ci/jobguard.sh
@@ -2,7 +2,7 @@
 
 set -e
 
-JOB_NAME_PATTERN=${JOB_NAME_PATTERN:-"(pre-main-kyma-components-.*)|(pre-main-kyma-tests-.*)|(pre-kyma-components-.*)|(pre-kyma-tests-.*)|(pull-.*-build)"}
+JOB_NAME_PATTERN=${JOB_NAME_PATTERN:-"(pull-.*-build)"}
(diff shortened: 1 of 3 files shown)


hack/ci/jobguard.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+set -e
+
+JOB_NAME_PATTERN=${JOB_NAME_PATTERN:-"(pre-main-kyma-components-.*)|(pre-main-kyma-tests-.*)|(pre-kyma-components-.*)|(pre-kyma-tests-.*)|(pull-.*-build)"}
```
triffer, line 5: Do we need the patterns "pre-main-kyma-components-.*", "pre-main-kyma-tests-.*", "pre-kyma-components-.*" and "pre-kyma-tests-.*"? As far as I understand we only use this as PR build guard for now and therefore "pull-.*-build" should be enough to avoid confusion.

barchw: Your right, I left that for consistency with API Gateway, but we should probably just remove it
```suggestion
JOB_NAME_PATTERN=${JOB_NAME_PATTERN:-"(pull-.*-build)"}
```

triffer (approved):

PR #177, opened 2023-06-13
Change CRD's IngressGateways to be a single object

### LINKED ISSUE #165: User is able to configure k8s resources for ingress gateway
**Description**

Atm we expose ingress gateway configuration as array but it does not match our default ingress gateway configuration in IstioOperator. 

ACs:
- [ ] it is possible to configure one ingress gateway

**Expected result**

It is possible to configure only one ingress gateway

**Actual result**

It is possible to configure array of ingress gateways

**Steps to reproduce**

install istio module with array of ingress gateways

kolodziejczak:
**Description**

- Change CRD's definition to accept one IngressGateway instead of array

**Related issue(s)**

Fixes #165

### DIFF
--- a/api/v1alpha1/istio_structs.go
+++ b/api/v1alpha1/istio_structs.go
@@ -16,7 +16,7 @@ type Components struct {
 	// Pilot defines component configuration for Istiod
 	Pilot *IstioComponent `json:"pilot,omitempty"`
 	// IngressGateways defines component configurations for Istio Ingress Gateways
-	IngressGateways *IstioComponent `json:"ingressGateways,omitempty"`
+	IngressGateway *IstioComponent `json:"ingressGateway,omitempty"`
 	// Cni defines component configuration for Istio CNI DaemonSet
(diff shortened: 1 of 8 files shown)


api/v1alpha1/istio_structs.go
```diff
@@ -16,7 +16,7 @@ type Components struct {
 	// Pilot defines component configuration for Istiod
 	Pilot *IstioComponent `json:"pilot,omitempty"`
 	// IngressGateways defines component configurations for Istio Ingress Gateways
-	IngressGateways []*IstioComponent `json:"ingressGateways,omitempty"`
+	IngressGateway *IstioComponent `json:"ingressGateway,omitempty"`
```
triffer, line 19: We need to update the documentation on this:
```
// IngressGateway defines component configurations for Istio Ingress Gateway
```

kolodziejczak: thanks, fixed

triffer (approved):

Issue #161, opened 2023-06-05
Istio CR stuck in Error state during uninstall


triffer:
**Description**

The Istio CR can be stuck in error state when the Istio uninstall was executed, but one of the steps after the uninstall (remove sidecars, remove finalizers) failed. In this case another call to uninstall will fail, since it's not idempotent.

**Expected result**
Istio CR is error state will recover from it during the next reconciliation and deleted from the cluster, even if Istio is not installed any longer on the cluster.

**Actual result**
Istio CR is stuck in error state and reconciliations fail with an error since Istio can't be uninstalled.

```
2023-06-05T06:44:48Z	INFO	Starting istio uninstall
2023-06-05T06:44:48Z	ERROR	Error occurred during reconciliation of Istio installation	{"error": "the server could not find the requested resource (get authorizationpolicies.security.istio.io)"}
github.com/kyma-project/istio/operator/controllers.(*IstioReconciler).Reconcile
	/workspace/controllers/istio_controller.go:79
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.2/pkg/internal/controller/controller.go:122
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.2/pkg/internal/controller/controller.go:323
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.2/pkg/internal/controller/controller.go:274
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.2/pkg/internal/controller/controller.go:235

```

**Steps to reproduce**
Manual using Istioctl:
This can be reproduced by having Istio installed by the istio operator. Then uninstall Istio using Istioctl and delete the Istio CR to trigger the uninstall.

Using integration tests:
This issue happens sometimes during Integration tests. In this case the removal of the finalizer fails.

- Remove `forceIstioCrRemoval` from `istioCrTearDown` hook in Istio integration tests
- Run one or multiple scenarios

**PRs**
- [x] https://github.com/kyma-project/istio/pull/178

**Troubleshooting**

PR #178, opened 2023-06-13
Fix for Istio uninstall when Istio already manually uninstalled

videlov:
**Description**

- Fixes Istio module uninstall if Istio CRDs/Resources removed from the cluster

**Related issue(s)**

https://github.com/kyma-project/istio/issues/161


internal/resources/resources.go
```diff
@@ -78,7 +79,7 @@ func (i *IstioResourcesFinder) FindUserCreatedIstioResources() ([]Resource, erro
 		u.SetGroupVersionKind(resource.GroupVersionKind)
 		err := i.client.List(i.ctx, &u)
 		if err != nil {
-			return nil, err
+			continue
```
barchw, line 82: Maybe some kind of assertion (e.g. k8serrors.IsNotFound) might be possible instead?
Or at least logging the error

tests/integration/scenario.go
```diff
@@ -29,4 +29,5 @@ func initScenario(ctx *godog.ScenarioContext) {
 	ctx.Step(`^Application "([^"]*)" in namespace "([^"]*)" has proxy with "([^"]*)" set to cpu - "([^"]*)" and memory - "([^"]*)"$`, steps.ApplicationHasProxyResourcesSetToCpuAndMemory)
 	ctx.Step(`^Application pod "([^"]*)" in namespace "([^"]*)" has Istio proxy "([^"]*)"$`, steps.ApplicationPodShouldHaveIstioProxy)
 	ctx.Step(`^Destination rule "([^"]*)" in namespace "([^"]*)" with host "([^"]*)" exists$`, steps.CreateDestinationRule)
+	ctx.Step(`^And Istio is manually uninstalled$`, steps.UninstallIstio)
```
triffer, line 32: The test steps are skipped since this step couldn't be found. Unfortunately, this doesn't cause an exit code unequal to 0 for the test run, yet. I'll push changes to fix this behaviour.
```suggestion
	ctx.Step(`^Istio is manually uninstalled$`, steps.UninstallIstio)
```

triffer (changes requested):
I think we also need to handle the is not found case for the [deletion of the istio-system](https://github.com/triffer/istio/blob/4e5955fdc7f9626cc731a100e1d7b07ca6a14245/internal/reconciliations/istio/client.go#L110-L110). This caused the integration test to fail for me locally with the error 
```
Error occurred during reconciliation of Istio installation	{"error": "namespaces \"istio-system\" not found"}
```

tests/integration/features/istio/production/istio_installation.feature
```diff
@@ -67,3 +67,13 @@ Feature: Installing and uninstalling Istio module
     Then "Istio CR" is not present on cluster
     And Istio CRDs "should not" be present on cluster
     And Namespace "istio-system" is "not present"
+
+Scenario: Uninstallation of Istio module if Istio was manually deleted
```
triffer, line 71: I again executed the scenario multiple times locally and I realized that it is flaky. 

The source of the flakiness is that sometimes the deletion of istio-system fails:
```
Error occurred during reconciliation of Istio installation	{"error": "namespaces \"istio-system\" not found"}
```

I think the steps in the current scenario are not correct. After uninstalling Istio manually we need to verify that istio-system and Istio CRDs are not present on the cluster any longer. Only then we can continue with triggering the uninstallation by the reconciler:

```
Scenario: Uninstallation of Istio module if Istio was manually deleted
    Given Istio CR "istio-sample" is applied in namespace "default"
    And Istio CR "istio-sample" in namespace "default" has status "Ready"
    And Namespace "istio-system" is "present"
    And Istio is manually uninstalled
    And Namespace "istio-system" is "not present"
    And Istio CRDs "should not" be present on cluster
    When "Istio CR" "istio-sample" in namespace "default" is deleted
    Then "Istio CR" is not present on cluster
```

Having the steps like that the scenario will constantly fail. We need to introduce the proposal from [my comment](https://github.com/kyma-project/istio/pull/178#pullrequestreview-1482743251) to fix it.

triffer (changes requested):

triffer (approved):

PR #190, opened 2023-06-27
Controller memory issue

### LINKED ISSUE #157: Controller memory issue
**Description**

After several reconciliations (for me ~3) that trigger a call to istio Install, the controller terminates because it hits the memory limit. It looks like there is a memory leak that causes the memory usage to increase with each reconciliation.

**Expected result**

Reconciliations can be executed without controller being OOM killed.

**Actual result**
Controller is OOM killed.

**Steps to reproduce**
- Deploy the controller
- Apply Istio CR
- Update Istio CR multiple times to trigger a call to Istio Installation

**Troubleshooting**

kolodziejczak:
**Description**

- Istio install execution in external process as a temporary hack for memory leak in Istio
- Integration tests check if the controller has been restarted

**Related issue(s)**

Fixes #157


internal/reconciliations/istio/client.go
```diff
@@ -119,6 +121,17 @@ func (c *IstioClient) Uninstall(ctx context.Context) error {
 	return nil
 }
 
+func installIstioInExternalProcess(mergedIstioOperatorPath string) (string, error) {
+	b, err := exec.Command("./istio_install", mergedIstioOperatorPath).Output()
```
triffer, line 125: Should we have a timeout for this? to not block the reconciliation loop in case of some unexpected issue?

triffer, line 125: With this approach we don't see the output of the Istio install while it's running, but only after the process has exited.
We should use the os.Stdout for the cmd, e.g. like this

```
	cmd := exec.Command("./istio_install", mergedIstioOperatorPath)
	cmd.Stdout = os.Stdout
	cmd.Stderr = os.Stderr
	err := cmd.Run()
```

cmd/istio-install/main.go
```diff
@@ -0,0 +1,43 @@
...
+
+func main() {
+	iopFileNames := []string{os.Args[1]}
+
+	istioLogOptions := initializeLog()
+	registeredScope := istiolog.RegisterScope("installation", "installation", 0)
+	consoleLogger := clog.NewConsoleLogger(os.Stdout, os.Stderr, registeredScope)
+	printer := istio.NewPrinterForWriter(os.Stdout)
+
+	// We don't want to verify after installation, because it is unreliable
```
videlov, line 36: we do it, just not here, but here:
https://github.com/kyma-project/istio/blob/6a2e6dca4c2de0100fb337043153e790fc14fce3/internal/reconciliations/istio/installation.go#L97-L104

tests/integration/hooks.go
```diff
@@ -42,6 +44,31 @@ var istioCrTearDown = func(ctx context.Context, sc *godog.Scenario, _ error) (co
...
+	err = c.List(ctx, podList, client.MatchingLabels{"app.kubernetes.io/component": "istio-operator.kyma-project.io"})
+	if err != nil {
+		return ctx, err
+	}
+	if len(podList.Items) < 1 {
+		return ctx, errors.New("Controller not found")
+	}
+
+	for _, cpod := range podList.Items {
+		if rc := cpod.Status.ContainerStatuses[0].RestartCount; rc > 0 {
```
videlov, line 63: did you check if this is the correct counter, e.g. did you check if you change the spec and container is restarted this is the place it gets incremented?

Makefile
```diff
@@ -115,10 +118,12 @@ test: manifests generate fmt vet envtest ## Run tests.
 .PHONY: build
 build: generate fmt vet ## Build manager binary.
 	go build -o bin/manager main.go
+	go build -o $(ISTIO_INSTALL_BIN_PATH) cmd/istio-install/main.go
 
 .PHONY: run
 run: manifests generate fmt vet ## Run a controller from your host.
-	go run ./main.go
+	go build -o $(ISTIO_INSTALL_BIN_PATH) cmd/istio-install/main.go
```
videlov, line 125: this line duplicates with 121, could we eventually move istio-install building in separate task?

videlov (changes requested):

cmd/istio-install/main.go
triffer: As far as I know this comment adds an information why we explicitly set the verify to false in the installation options, so I think it's valid.

videlov: ahhh, I see, comment it did mention verify parameter, I thought it is a general statement.

Makefile
kolodziejczak: Thanks for pointing it out.

tests/integration/hooks.go
videlov: thats not good, what if container crashes, we need to be sure this is the right counter in case container crashes. Will try it out on my instance just to make sure.

kolodziejczak: In addition to OOM it will check also for other potential issues that might happen to the controller and crash it like eg. runtime errors.

Makefile
kolodziejczak: Changed

tests/integration/hooks.go
videlov: this is the correct counter.

videlov (approved):

Issue #181, opened 2023-06-19
Time based reconcile


strekm:
**Description**

At this moment istio controller reacts only on changes on Istio CR. Implement time based reconciliation that instals istio. In first iteration default time reconciliation period is ok, we can adjust later if needed. Expose this configuration so this configuration can be adjusted without changes in source code.

- [x] Istio is reconciled on timely manner
- [x] verify `service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout` annotation are reset to default on ingressgateway service
- [x] Since we'll touch Change evaluation of Istio CR, we should also be able to fix the issue https://github.com/kyma-project/istio/issues/150

**DoD:**
- [x] provide documentation
~~- [ ] release notes and What's New updates for Kyma customers~~
- [x] provide unit tests
- [x] provide integration tests
- [x] test on production-like environment
- [x] verify resource limits
- [x] PR reviewer will verify code coverage and evaluate if it is acceptable

**Attachments**
part of: https://github.com/kyma-project/istio/issues/38

PR #195, opened 2023-07-04
Time based reconcile

werdes72:
**Description**

- Add configurable reconcile interval
- Run Istio install even when there are no CR changes
- Update go.mod

See: #181


internal/reconciliations/istio/installation.go
```diff
@@ -33,27 +33,22 @@ const (
...
 		if err := i.Client.Update(ctx, &istioCR); err != nil {
 			return istioCR, err
 		}
 	}
 
-	ctrl.Log.Info("Reconcile Istio installation", "Istio CR change evaluation", istioCRChanges)
-
-	if istioCRChanges.requireInstall() {
-
+	if shouldInstallIstio {
```
triffer, line 51: I think we don't need this separation between Install and deletion. It might be confusing. What other cases than install and Uninstall are there? Since we only have deletion and "default handling" (call Istio install) at the moment. We can some conditions and make the code flow easier to follow.
So I think it might be more clear if we have something like
```
if shouldDelete
 Deletion logic
else
 Logic verifying Istio Version (currently in shouldInstall)
 Add finalizer if necessary
 Execute install logic
```

What do you think?

triffer: So if we want to keep the shouldInstallIstio, then for me it makes more sense to have the initial positive action (the install) first and then the deletion as the else-if case. But I still don#t see the usage of an `else if` since we don't expect any different case.

```
if shouldInstall
  Add finalizer if needed
  Execute install logic
else if shouldDelete
  Deletion logics
```

main.go
```diff
@@ -43,6 +43,7 @@ const (
 	rateLimiterFrequencyDefault = 30
 	failureBaseDelayDefault     = 1 * time.Second
 	failureMaxDelayDefault      = 1000 * time.Second
+	retryTimeDefault            = 10 * time.Hour
```
triffer, line 46: Should we call this `requeueAfterTime`, since retry indicates that something failed and I need to try it again, but this is just a time that will also be used for non-error scenarios. Another idea is `reconciliationInterval`?

triffer (approved):

main.go
```diff
@@ -134,6 +136,8 @@ func defineFlagVar() *FlagVar {
 	flag.DurationVar(&flagVar.failureBaseDelay, "failure-base-delay", failureBaseDelayDefault,
 		"Indicates the failure base delay in seconds for rate limiter.")
 	flag.DurationVar(&flagVar.failureMaxDelay, "failure-max-delay", failureMaxDelayDefault,
-		"Indicates the failure max delay in seconds")
+		"Indicates the failure max delay in seconds.")
+	flag.DurationVar(&flagVar.retryTime, "retry-time", retryTimeDefault,
+		"Indicates the time based reconciliation interval in seconds.")
```
triffer, line 141: This info might be misleading, since it's not in seconds but a duration. So `30` is an invalid input, but `30s`, `3m` or `4h` is valid.

internal/reconciliations/istio/installation.go
werdes72: There is a different case - when delete action has been already called. In this case we won't execute install nor delete logic. It's also covered in tests.

triffer (approved):

PR #206, opened 2023-07-06
Add RNs for 0.1 version

nataliasitko:

- Add RNs for version 0.1


kolodziejczak (dismissed):

barchw (commented):
File name should follow semantic versioning `0.1.0.md`

kolodziejczak:
Yeah, sorry for misinforming. It would still pass with the wrong format so maybe we should provide some format validation to the workflow scripts?

barchw:
> Yeah, sorry for misinforming. It would still pass with the wrong format so maybe we should provide some format validation to the workflow scripts?

Release will fail if set to `0.1.0`. I think we cannot really validate the naming of RN file before, unless it is a separate workflow triggered on creation e.g. linting

docs/release-notes/0.1.0.md
```diff
@@ -0,0 +1 @@
+We are pleased to present the preliminary version of the Istio module. This release is primarily meant for internal testing purposes. The official release is expected to be available in about one week.
```
mmitoraj, line 1:
```suggestion
We are pleased to present the preliminary version of the Istio module. This release is primarily meant for internal testing purposes. Expect the official release in about one week.
```

mmitoraj (approved):

PR #207, opened 2023-07-06
Disallow non semantic versioned tags

barchw:
**Description**

- Allow to create releases only with semantic versioned tags

### DIFF
--- a/scripts/create_draft_release.sh
+++ b/scripts/create_draft_release.sh
@@ -10,7 +10,7 @@ set -o pipefail # prevents errors in a pipeline from being masked
 
 RELEASE_TAG=$1
 
-if ! [[ "$RELEASE_TAG" =~ ^[0-9]\.[0-9]\.[0-9](-.+)?$ ]]; then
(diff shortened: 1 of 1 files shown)


scripts/create_draft_release.sh
```diff
@@ -10,6 +10,11 @@ set -o pipefail # prevents errors in a pipeline from being masked
 
 RELEASE_TAG=$1
 
+if ! [[ "$RELEASE_TAG" =~ ^[0-9]\.[0-9]\.[0-9](-.+)?$ ]]; then
```
triffer, line 13: I think we won't be able to release any major, minor or patch version that is equal to or greater than 10. 
I think at some point this will hit us.

barchw: Ah your right, let me adjust that

barchw:
```suggestion
if ! [[ "$RELEASE_TAG" =~ ^[0-9]+\.[0-9]+\.[0-9]+(-.+)?$ ]]; then
```

triffer (approved):

Issue #193, opened 2023-06-29
Ingress gateway restart after numTrustedProxy change


triffer:
**Description**

In the old reconciler there is an implementation, that after the change of the numTrustedProxy configuration in the Istio CR, the Istio ingress gateway is restarted. This needs to be done so that the new configuration is picked up by the ingress gateway, since calling Istio install won't restart the ingress gateway.
The [original issue](https://github.com/kyma-incubator/reconciler/issues/1246) to implement this in the reconciler has some additional infos.

**Reasons**
Istio module must support the same behaviour as the old reconciler.

**DoD:**
- [x] provide documentation
- [x] provide unit tests
- [x] provide integration tests
- [x] test on production-like environment
- [x] PR reviewer will verify code coverage and evaluate if it is acceptable

**PR:**
- [x] https://github.com/kyma-project/istio/pull/208

**Attachments**

PR #208, opened 2023-07-06
Restart Istio GW on changes of numTrustedProxies configuration

videlov:
**Description**

- Restart Istio GW on changes of numTrustedProxies configuration

**Related issue(s)**

https://github.com/kyma-project/istio/issues/193


internal/ingress-gateway/config.go
```diff
@@ -0,0 +1,57 @@
...
+
+const (
+	configMapNS   string = "istio-system"
+	configMapName string = "istio"
+)
+
+// GetNumTrustedProxyFromIstioCM fetches current cluster configuration for "numTrustedProxies" from istio-system/istio ConfigMap.
+// Returns nil if no configuration is present (default configuration).
+// The desired way to get the configuration would be to fetch it from Istio Operator, but at this point in time
+// it doesn't reflect the actual state of the cluster. This should change when the functionality will be migrated
```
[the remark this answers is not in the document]
videlov: Adapted to read config from `lastAppliedConfig` annotation.

internal/reconciliations/istio/configuration_test.go
```diff
@@ -1,32 +1,180 @@
...
+var _ = Describe("Ingress Gateway", func() {
+	Context("IngressGatewayNeedsRestart", func() {
+		It("should restart when CR numTrustedProxies is 2 and in lastAppliedConfig is 1", func() {
+			//given
+			oldNumTrustedProxies := 1
+
+			istioCR := operatorv1alpha1.Istio{}
+			istioCR.Spec.Config.NumTrustedProxies = &oldNumTrustedProxies
+
+			updatedCR, err := istio.UpdateLastAppliedConfiguration(istioCR, mockIstioTag)
```
triffer, line 59: We use the function `UpdateLastAppliedConfiguration` to setup test data without having an explicit test that is testing its working as expected. I think we should have a test verifying the correct behaviour of the function if we want to use it to setup test data. I personally prefer to not rely on other implementation, but create static test data for the test to avoid failing tests, because of bugs in functionality that is not the subject under test.

docs/contributor/04-10-technical-design.md
```diff
@@ -94,54 +107,63 @@ in order to trigger them at different intervals, or to have a different interval
...
+
 The PeerAuthenticationReconcilation component applies the PeerAuthentication that configures default mTLS mode in a cluster.
 The PeerAuthenticationReconcilation must only be applied if Istio is installed and PeerAuthentication does not already exist or the generation is changed, as we want to ensure that it is always our expected configuration.
 
 ## Istio component uninstallation
 
 The default behaviour triggered on deletion of Istio Custom Resource is to uninstall all of Istio components only if there are none customer created resources present on the cluster. This behaviour is called `blocking` deletion strategy and will take place unless the intent to delete all resources, including non default Istio ones, is explicitly defined by selecting `cascading` deletion strategy.
 
-> TODO: At this moment only `blocking` strategy is implemented and triggered by default. Implement `cascading` strategy as described in https://github.com/kyma-project/istio/issues/130
+> TODO: At this moment only `blocking` strategy is implemented and triggered by default. Implement `cascading` strategy as described in <https://github.com/kyma-project/istio/issues/130>
```
nataliasitko, line 164:
```suggestion
> TODO: At this moment only `blocking` strategy is implemented and triggered by default. Implement `cascading` strategy as described in this [issue](https://github.com/kyma-project/istio/issues/130).
```

internal/reconciliations/istio/installation.go
```diff
@@ -82,6 +84,11 @@ func (i *Installation) Reconcile(ctx context.Context, istioCR operatorv1alpha1.I
 			return istioCR, described_errors.NewDescribedError(err, "Could not get configuration from Istio Operator file")
 		}
 
+		ingressGatewayNeedsRestart, err := IngressGatewayNeedsRestart(istioCR)
```
triffer, line 87: Right now the invocation of ` IngressGatewayNeedsRestart` and `RestartIngressGateway` are two separate function calls that are also a little bit separated in the code. I think now hat we rely on lastAppliedConfiguration instead of the CM on the cluster, we can also encapsulate the IngressGatewayNeedsRestart in the function doing the IGW restart. This will make the code in the installation process easier and we can have the logic about the ingress gateway restart more encapsulated.

pkg/lib/common/common.go
triffer, line 1: Can we try to give it some better name than common in common? This will just mean we will add more things into a common package without providing any scope. We could put it in a package `annotations` and a `restart.go` file.

pkg/lib/common/common.go
```diff
@@ -0,0 +1,23 @@
...
+func AddRestartAnnotation(annotations map[string]string) map[string]string {
+	if len(annotations) == 0 {
+		annotations = map[string]string{}
+	}
+
+	annotations[restartAnnotationName] = time.Now().Format(time.RFC3339)
+	return annotations
+}
+
+func WasRestarted(annotations map[string]string) bool {
```
triffer, line 20: Maybe we should call this `HasRestart`, since the annotations itself can't be restarted and it aligns better with the `AddRestartAnnotation`.

pkg/lib/common/common_test.go
```diff
@@ -0,0 +1,53 @@
...
+	})
+})
+
+var _ = Describe("WasRestarted", func() {
+	It("should return false when restart annotation is missing", func() {
+		// when
+		wasRestarted := common.WasRestarted(nil)
+
+		// then
+		Expect(wasRestarted).To(BeTrue())
```
triffer, line 40: Test says that it should return false, but the expect tests for true.

internal/reconciliations/istio/installation.go
```diff
@@ -82,6 +85,11 @@ func (i *Installation) Reconcile(ctx context.Context, istioCR operatorv1alpha1.I
 			return istioCR, described_errors.NewDescribedError(err, "Could not get configuration from Istio Operator file")
 		}
 
+		ingressGatewayNeedsRestart, err := ingressgateway.NeedsRestart(ctx, i.Client, &istioCR)
```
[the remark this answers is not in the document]
triffer: 👍 Good job, this should work. The only downside I can see is, that we don't rely on the real state of the Istio resources, but our assumption that is stored as last applied config in the Istio CR. But since we can't easily find out what's the current configuration of the IGWs, it's better to use our last applied annotation than Istios config map. Still a user could modify it and we'd not be able to detect it, but I'd consider this an absolut corner case.

internal/reconciliations/istio/configuration.go
```diff
@@ -86,6 +113,34 @@ func CheckIstioVersion(currentIstioVersionString, targetIstioVersionString strin
 	return nil
 }
 
+func IngressGatewayNeedsRestart(istioCR operatorv1alpha1.Istio) (bool, error) {
```
triffer, line 116: Do we want/need to expose this?

internal/reconciliations/istio/configuration.go
```diff
@@ -86,6 +113,34 @@ func CheckIstioVersion(currentIstioVersionString, targetIstioVersionString strin
...
+		return true, nil
+	}
+	if isNewNotNil != isOldNotNil {
+		return true, nil
+	}
+
+	return false, nil
+}
+
+func RestartIngressGateway(ctx context.Context, k8sClient client.Client) error {
```
triffer, line 134: Do we want/need to expose this?

internal/reconciliations/istio/installation_test.go
triffer, line 1: We don't have negative tests to verify that installation will return an error if IngressGateway couldn't be restarted. Do want to have this tests?

pkg/lib/common/common_test.go
triffer, line 1: This tests won't be executed, since the Ginkgo initialization is missing, e.g. [here](https://github.com/kyma-project/istio/blob/869d96f1559491ed514e0b129015b0d56fa6cc82/pkg/lib/gatherer/gatherer_test.go#L37). Also the junit report generation must be added like [here](https://github.com/kyma-project/istio/blob/869d96f1559491ed514e0b129015b0d56fa6cc82/pkg/lib/gatherer/gatherer_test.go#L43).

internal/reconciliations/istio/configuration.go
```diff
@@ -1,34 +1,45 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	ingressgatewayNamespace      string = "istio-system"
+	ingressgatewayDeploymentName string = "istio-ingressgateway"
 )
 
-type appliedConfig struct {
+type AppliedConfig struct {
```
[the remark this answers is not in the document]
videlov: Correct.

internal/reconciliations/istio/configuration_test.go
videlov: Function is tested in the first test in Context `LastAppliedConfiguration`, but I can agree it is not so nice to reuse other code to setup tests. I just wanted to avoid some code/func duplication. Will redo it.

internal/reconciliations/istio/configuration.go
videlov: I did refactor it so its a single method now, but unfortunately we need to export it because it is used by the tests that are in `xxx_test` package.

videlov: did refactor test file to use same package as the implementation, as we discussed.

internal/reconciliations/istio/configuration.go
videlov: same.

pkg/lib/common/common.go
videlov: agree to reduce scope from `common` to `annotations`, common is too general

triffer (approved):

nataliasitko (approved):

PR #209, opened 2023-07-07
Warning status

werdes72:
**Description**

- Add warning status to Istio module

**Related issue(s)**
https://github.com/kyma-project/lifecycle-manager/issues/525

### DIFF
--- a/docs/user/01-20-istio-custom-resource.md
+++ b/docs/user/01-20-istio-custom-resource.md
@@ -37,6 +37,6 @@ This table lists all the possible parameters of the given resource together with
 
 **Status:**
 
-| Parameter | Type | Description |
-| ---- | ----------- | ---- |
-| **state** (required) | string | Signifies the current state of **CustomObject**. Its value can be either `Ready`, `Processing`, `Error`, or `Deleting`. |
+| Parameter | Type | Description                                                                                                                       |
+| ---- | ----------- |-----------------------------------------------------------------------------------------------------------------------------------|
+| **state** (required) | string | Signifies the current state of **CustomObject**. Its value can be either `Ready`, `Processing`, `Error`, `Warning` or `Deleting`. |
--- a/docs/user/01-20-istio-custom-resource.md
+++ b/docs/user/01-20-istio-custom-resource.md
@@ -39,4 +39,4 @@ This table lists all the possible parameters of the given resource together with
 
 | Parameter | Type | Description                                                                                                                       |
 | ---- | ----------- |-----------------------------------------------------------------------------------------------------------------------------------|
-| **state** (required) | string | Signifies the current state of **CustomObject**. Its value can be either `Ready`, `Processing`, `Error`, `Warning` or `Deleting`. |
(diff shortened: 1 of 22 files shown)


internal/status/status.go
```diff
@@ -60,3 +61,9 @@ func (DefaultStatusHandler) SetDeleting(ctx context.Context, client client.Clien
 	istioCR.Status.State = operatorv1alpha1.Deleting
 	return update(ctx, client, istioCR, condition, retryTime...)
 }
+
+func (DefaultStatusHandler) SetWarning(ctx context.Context, err described_errors.DescribedError, client client.Client, istioCR *operatorv1alpha1.Istio, condition metav1.Condition, retryTime ...time.Duration) (ctrl.Result, error) {
```
barchw, line 65: As warning is also a `described_error` I think that moving the logic under `SetError` func would make more sense (so we should set `istioCR.Status.State` based on err.level()). Then the new logic in https://github.com/kyma-project/istio/blob/fa38c94d0718c4153f1d2adf55452206d3b0edba/controllers/istio_controller.go#L84 could be removed, only leaving SetError at the end.

internal/reconciliations/proxy/sidecars_test.go
```diff
@@ -141,3 +141,7 @@ func (s StatusMock) SetError(_ context.Context, _ described_errors.DescribedErro
 func (s StatusMock) SetDeleting(_ context.Context, _ client.Client, _ *operatorv1alpha1.Istio, _ metav1.Condition, _ ...time.Duration) (ctrl.Result, error) {
 	return s.result, s.err
 }
+
```
barchw, line 144: Maybe the name for `described_errors.DescribedError` should be adjusted (e.g. `described_errors.DescribedState`)?

barchw (changes requested):
Unit test coverage is missing

barchw (dismissed):

docs/user/01-20-istio-custom-resource.md
```diff
@@ -37,6 +37,6 @@ This table lists all the possible parameters of the given resource together with
 
 **Status:**
 
-| Parameter | Type | Description |
-| ---- | ----------- | ---- |
-| **state** (required) | string | Signifies the current state of **CustomObject**. Its value can be either `Ready`, `Processing`, `Error`, or `Deleting`. |
\ No newline at end of file
+| Parameter | Type | Description                                                                                                                       |
+| ---- | ----------- |-----------------------------------------------------------------------------------------------------------------------------------|
+| **state** (required) | string | Signifies the current state of **CustomObject**. Its value can be either `Ready`, `Processing`, `Error`, `Warning` or `Deleting`. |
```
nataliasitko, line 42:
```suggestion
| **state** (required) | string | Signifies the current state of **CustomObject**. Its value can be either `Ready`, `Processing`, `Error`, `Warning`, or `Deleting`. |
```

nataliasitko (approved):

barchw (approved):

PR #210, opened 2023-07-07
Kubernetes compatibility GHA

barchw:
**Description**

- Add Kubernetes compatibility GHA


.github/workflows/pull-request.yaml
```diff
@@ -53,4 +56,78 @@ jobs:
...
+  changes:
+    runs-on: ubuntu-latest
+    # Required permissions
+    permissions:
+      pull-requests: read
+    # Set job outputs to values from filter step
+    outputs:
+      change: ${{ steps.filter.outputs.api == 'true' || steps.filter.outputs.cmd == 'true' || steps.filter.outputs.config == 'true' || steps.filter.outputs.controllers == 'true' || steps.filter.outputs.internal == 'true' || steps.filter.outputs.manifests == 'true' || steps.filter.outputs.pkg == 'true' || steps.filter.outputs.main == 'true' || steps.filter.outputs.mod == 'true' }}
+    steps:
+      - uses: dorny/paths-filter@4512585405083f25c027a35db413c2b3b9006d50 # v2.11.1
```
triffer, line 70: I don't know if we should use a separate 3rd party action for this. I think we can achieve the same when we create a new workflow that runs on PRs and uses [path filter](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-paths).

barchw: Yeah good idea. I was looking for this in GH documentation but couldn't find :)

.github/workflows/pull-request.yaml
```diff
@@ -27,8 +27,10 @@ jobs:
           allowlist: |
             actions/checkout
             actions/setup-go
+            actions/upload-artifact
 
   lint:
+    needs: [verify-actions]
```
triffer, line 33: Instead of creating a dependency between verify-actions and the other jobs, I thought about moving this into a separate workflow that only runs if a GitHub workflow changed, since this is the only trigger relevant for this.

.github/workflows/pull-request.yaml
barchw: https://github.com/kyma-project/istio/pull/210/commits/c5a78ac61fa296ad7748674fe153c471a794cd0d

.github/workflows/pull-request.yaml
barchw: https://github.com/kyma-project/istio/pull/210/commits/8ee882fd6e61b7c9e955cc5b5fbb82e4a82dcdff

PR #222, opened 2023-07-11
GitHub Actions cache build on main branch

triffer:
**Description**
With our current config we create a cache with go dependencies for every PR. This caches can only be re-used in the same PR. That means the first workflows in a PR always have to download the dependencies, even when there are not changes to the dependencies.
To avoid that we can leverage the built-in usage of caches from the main branch. 

> Access restrictions provide cache isolation and security by creating a logical boundary between different branches or tags. Workflow runs can restore caches created in either the current branch or the default branch (usually main).  

[GitHub cache docs](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows)

- Add main workflow with a step to run go build and create the cache

**Related issue(s)**

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -1,0 +1,30 @@
+name: Main
+
+on:
+  push:
+    branches:
+      - main
+    paths-ignore:
+      - 'docs/**'
+      - '**/*.md'
+      - 'tests/performance/**'
+      - 'OWNERS'
+      - 'CODEOWNERS'
+      - 'sec-scanners-config.yaml'
+      - './github/dependabot.yml'
+permissions:
+  contents: read
+
+jobs:
+
+  run-unit-tests:
+    name: Build cache
+    description: Build cache on main so it can be reused in branches.
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v4
+        with:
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yaml
```diff
@@ -0,0 +1,28 @@
+name: Main
+
+on:
+  push:
+    branches:
+      - main
+    paths-ignore:
+      - 'docs/**'
+      - '**/*.md'
+      - 'tests/performance/**'
```
barchw, line 10: Maybe we should also cache on changes in perf tests
```suggestion
```

barchw (changes requested):

.github/workflows/main.yaml
triffer: Since this only applies to caching Go dependencies, I don't see a benefit of running this on performance test changes as well, or am I missing something?

.github/workflows/main.yaml
```diff
@@ -0,0 +1,28 @@
...
+      - 'OWNERS'
+      - 'CODEOWNERS'
+      - 'sec-scanners-config.yaml'
+      - './github/dependabot.yml'
+
+jobs:
+
+  run-unit-tests:
```
triffer, line 18: Fix naming of job

barchw (approved):

Issue #192, opened 2023-06-27
Object already has been modified error during reconciliation


kolodziejczak:
**Description**
During Istio reconciliation the error occurs which indicates that the object we are trying to change has already been modified.
It happens after update to IstioCR
It is most likely due to a stale cache.

**Expected result**
Reconciliation without "object has been modified" error

**Actual result**
![istio_cr_stale_object](https://github.com/kyma-project/istio/assets/69915024/2b307142-beb3-4150-96a0-29038387eca2)

**Steps to reproduce**
1. Apply IstioCR 
2. Update the same IstioCR

PR #253, opened 2023-07-19
Fix object modified errors

triffer:
**Description**

- Update the correct status as early as possible in the controller instead of the doing this in the sub reconciliations. I think the controller is the better level to handle state updates for the CR. We can also think about moving the status handling to the controller package.
- Add retry on conflict for update of status, lastAppliedConfiguration annotation and finalizer removal since those have been a constant source of this issue and we can't avoid it.
- Avoid technical "exit status 1" error description in Istio CR when Istio installation in external process failed.
- Removed result handling from Status handling to make its responsibility more clear.
- Remove ErrorRetryTime to use the `DefaultControllerRateLimiter` behaviour instead of a linear retry time.

**Related issue(s)**

- https://github.com/kyma-project/istio/issues/192


kolodziejczak (commented):
reviewed

controllers/istio_controller_test.go
```diff
@@ -59,21 +64,416 @@ var _ = Describe("Istio Controller", func() {
...
+	return fake.NewClientBuilder().WithScheme(getTestScheme()).WithObjects(objects...).Build()
+}
+
+func getTestScheme() *runtime.Scheme {
+	Expect(operatorv1alpha1.AddToScheme(scheme.Scheme)).Should(Succeed())
+
+	return scheme.Scheme
+}
+
+type istioInstallationReconciliationMock struct {
```
videlov, line 434: did we consider using a mock framework for these generations, e.g. counterfeit, mockery, gomock etc ...

videlov (changes requested):

controllers/istio_controller.go
```diff
@@ -103,26 +113,65 @@ func (r *IstioReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
...
-	// Put applied configuration in annotation
-	istioCR, updateErr := istio.UpdateLastAppliedConfiguration(istioCR, IstioTag)
-	if updateErr != nil {
-		r.log.Error(updateErr, "Error updating LastAppliedConfiguration")
-		return r.statusHandler.SetError(ctx, described_errors.NewDescribedError(updateErr, "Error updating LastAppliedConfiguration"), r.Client, &istioCR, metav1.Condition{}, ErrorRetryTime)
+	return r.finishReconcile(ctx, istioCR, IstioTag)
+}
+
+// cancelReconcile cancels the reconciliation and requeues the request.
+func (r *IstioReconciler) cancelReconcile(ctx context.Context, istioCR operatorv1alpha1.Istio, err described_errors.DescribedError) (ctrl.Result, error) {
```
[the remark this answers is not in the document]
triffer: I created different functions to make it to make it easy to spot that those are different actions. Also the logic inside will get harder to understand, since there is a different handling for the result in successful and status error case and different logging. I'm not a fan of mixing this up, because this are completely different processes and we should make it obvious.

controllers/istio_controller_test.go
triffer: I think we were using some of those in the reconciler and I remember that people didn't like it. 
I'm also not a big fan of this mock implementations I did here, but also most Go mocking frameworks I saw are just so cumbersome with all that generated code ... I'll check how we can replace this.

triffer: After discussion we agreed to keep it as it is for now and see how things develop. If it gets more complex we should replace it.

videlov (approved):

PR #255, opened 2023-07-20
Wait for ResourceQuota status on perf test script

videlov:
**Description**

- Waits for ResourceQuota status on perf test script (with retries)

**Related issue(s)**


triffer:
We can close this, because we are no longer using resource quota.

PR #288, opened 2023-07-31
Refactor gh actions

kolodziejczak:
**Description**

- Refactor github actions

**Related issue(s)**


.github/workflows/post-integration.yaml
```diff
@@ -0,0 +1,58 @@
+# This workflow is resposible for all kinds of integration tests that are run AFTER MERGE TO MAIN
+# Those tests depend on a container image so in the first job we wait for the image build to succeed
+
+name: Post submit integration tests
```
triffer, line 4: Maybe we can just name them main integration tests, since this "post submit" convention is introduced by Prow. We can then also name the file `main-Integration.yaml`.

.github/workflows/post-integration.yaml
```diff
@@ -0,0 +1,58 @@
...
+      - main
+
+permissions:
+  contents: read
+
+jobs:
+  wait-for-image-build:
+    name: Wait for image build
+    runs-on: ubuntu-latest
+    if: github.event.pull_request.draft == false
```
triffer, line 26: I think we can remove this if. Since this is the main workflow.

.github/workflows/pull-integration.yaml
```diff
@@ -1,4 +1,4 @@
-# Jobs defined in this file are resposible for all kinds of integration tests
+# This workflow is resposible for all kinds of integration tests run on PULL REQUEST
```
triffer, line 1: Typo and I don't think we need to scream the pull request 😄 
```suggestion
# This workflow is responsible for all kinds of integration tests run on pull requests.
```

.github/actions/upgrade-integration-test/action.yaml
```diff
@@ -0,0 +1,25 @@
+# Runs upgrade suite of integration tests
+# Requires as input IMG ENV variable for make target
+# Don't forget to use actions/checkout@v3 and actions/setup-go@v4 before
+
+name: 'Upgrade integration test'
+description: 'Runs upgrade suite of integration tests'
+inputs:
+  img-url:
+    description: 'Env variable required for make target'
```
triffer, line 9: Might be also good to have a description what img-url is, you only added what it is used for. We can also think about adding an example to make it clear.

.github/actions/upgrade-integration-test/action.yaml
```diff
@@ -0,0 +1,25 @@
+# Runs upgrade suite of integration tests
+# Requires as input IMG ENV variable for make target
```
triffer, line 2: Do we need to have this as description, since it's already clearly visible in the action definition?

.github/actions/k8s-compatibility-test/action.yaml
```diff
@@ -0,0 +1,46 @@
+# Runs Kubernetes version compatibility check
+# Requires as input IMG ENV variable for make target
```
triffer, line 2: Do we need to have this as description, since it's already clearly visible in the action definition?

.github/actions/wait-for-job-succeed-or-fail/action.yaml
```diff
@@ -0,0 +1,32 @@
+# Waits for a given job to finish
+# On succees it passes, on failure it fails the job
+# inputs: job-name (required), timeout (optional)
+# Don't forget to use actions/checkout@v3 and actions/setup-go@v4 before
```
triffer, line 4: I think it should be possible to move the setup-go action also inside of the composite actions as step? This way the actions would be more self-contained.

```
runs:
  using: "composite"
  steps:
    - uses: actions/setup-go@v4
       with:
         go-version: "1.20"
    - name: Wait for job to succeed
```

triffer:
Since this PR contains only changes to GHA config files, the PR integration test workflow is not triggered, maybe we should think about also triggering Prow builds, so we can run the PR integration tests, to verify changes to the PR workflows.

.github/workflows/main-integration.yaml
```diff
@@ -0,0 +1,51 @@
+# This workflow is responsible for all kinds of integration tests that are run after push to main.
+# Those tests depend on a container image so in the first job we wait for the image build to succeed.
+
+name: Post submit integration tests
```
triffer, line 4:
```suggestion
name: main integration tests
```

triffer (approved):

PR #383, opened 2023-09-12
Add Integration test to verify access between namespaces

triffer:
**Description**

- Add integration test to verify connection between namespaces is working

**Related issue(s)**

### DIFF
--- a/tests/integration/steps/nginx.go
+++ b/tests/integration/steps/nginx.go
@@ -32,7 +32,7 @@ func CreateNginxApplication(ctx context.Context, appName, namespace, forwardTo s
 		return ctx, err
 	}
 
-	ctx, err = createService(ctx, appName, namespace, err)
+	ctx, err = createService(ctx, appName, namespace)
 	if err != nil {
 		return ctx, err
 	}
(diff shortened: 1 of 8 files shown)


tests/integration/steps/nginx.go
```diff
@@ -0,0 +1,198 @@
...
+			Namespace: namespace,
+		},
+		Spec: corev1.ServiceSpec{
+			Selector: map[string]string{
+				"app": appName,
+			},
+			Ports: []corev1.ServicePort{
+				{
+					Name:       "http",
+					Port:       80,
```
barchw, line 60: The port could also be dynamic

barchw: Unless we are making it so because the config map is hardcoded, then I would mention it as a hardcoded value in the feature as well

triffer: I agree, but I wanted to keep it simple for a starter. I think we can make it dynamic once we see a use case for this, e.g. how we did it with http bin application. Also I think for simplicity we might always want to have it as 80, since this is the most straight forward when doing http requests.

triffer: I added the the static port to the step

barchw (approved):

PR #443, opened 2023-10-26
Add Client IP in XFF header for AWS

kolodziejczak:
**Description**

- Add aws load balancer annotations
- Add envoy filters for AWS


internal/clusterconfig/aws_client.go
```diff
@@ -0,0 +1,31 @@
...
+type AwsClientInterface interface {
+	IsAws() bool
+}
+type AwsClient struct {
+	*http.Client
+}
+
+func NewAwsClient() *AwsClient {
+	return &AwsClient{
+		&http.Client{Timeout: 5 * time.Second},
```
triffer, line 21: The timeout is rather long, because on all non-AWS environments this will run into a timeout. So I think a rather short timeout like 1s or 2s would be better.

internal/clusterconfig/aws_client.go
```diff
@@ -0,0 +1,31 @@
+package clusterconfig
+
+import (
+	"net/http"
+	"time"
+)
+
+const awsMetadataHost = "http://169.254.169.254"
```
triffer, line 8: Only using this seems to be not reliable, since also other cloud provider use it, e.g. Azure: https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service

kolodziejczak: I'll add to the ip the exact http endpoint which will differentiate between providers. Eg /latest/meta-data will not be available outside of AWS.

internal/clusterconfig/hyperscaler_test.go
```diff
@@ -0,0 +1,51 @@
...
+				ts.URL,
+			}
+
+			// when
+			isAws := hc.IsAws()
+
+			// then
+			Expect(isAws).To(BeFalse())
+		})
+
```
videlov, line 49: there is no test when `http.Get()` returns an error:
```
r, err := hs.HttpClient.Get(hs.AwsMetadataHost)
	if err != nil {
        ...
```
if it is not crazy difficult, please include it as well.

internal/resources/disclaimer_annotation.go
```diff
@@ -0,0 +1,34 @@
...
+		annotations = make(map[string]string)
+	}
+	annotations[DisclaimerKey] = DisclaimerValue
+	resource.SetAnnotations(annotations)
+
+	err := k8sClient.Update(ctx, &resource)
+	return err
+}
+
+func hasManagedByDisclaimer(resource unstructured.Unstructured) bool {
```
videlov, line 27: can you please test this function as well (in `disclaimer_annotation_test.go`). you can utilize it in `apply_test.go` as well.

videlov (changes requested):

internal/clusterconfig/hyperscaler.go
```diff
@@ -0,0 +1,34 @@
+package clusterconfig
+
+import (
+	"net/http"
+)
+
+const AwsMetadataHost = "http://169.254.169.254/latest/meta-data/"
```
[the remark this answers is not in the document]
kolodziejczak: Yes, the paths differentiate between hyperscalers. Manual tests confirmed that it works as intended.

videlov (dismissed):

nataliasitko (approved):

videlov (approved):

PR #454, opened 2023-11-03
Integration tests gardener

kolodziejczak:
**Description**

- Scripts for provisioning gardener and running integration tests


tests/integration/scripts/integration-test.sh
```diff
@@ -0,0 +1,63 @@
...
+export PATH="${PATH}:${PWD}"
+
+CLUSTER_NAME=$(LC_ALL=C tr -dc 'a-z' < /dev/urandom | head -c10)
+export CLUSTER_NAME
+
+./tests/integration/scripts/provison-gardener.sh
+./tests/integration/scripts/jobguard.sh
+
+# this make target should have deploy as a prerequisite target
+make "$1"
```
triffer, line 63: It might be better to validate that $1 is passed at the beginning of the script and fail early in this case
```shell
if [ -z "$1" ]; then
    echo "Make target to run integration tests is required as first parameter, but not provided"
    exit 2
fi
```

tests/integration/scripts/integration-test.sh
triffer, line 1: Since this is specific for gardener, it might be better to name the file `integration-test-gardener.sh`.

tests/integration/scripts/integration-test.sh
kolodziejczak: right

triffer (approved):

PR #506, opened 2023-12-05
Remove label from default Istio CR

triffer:
**Description**

- Remove unnecessary label from default Istio CR

**Related issue(s)**


triffer:
Closing, because we will implement default labels that should be set by all Kyma modules

PR #551, opened 2024-01-05
Add tcp keepalive to default mesh config

Rand0mF:
This adds tcp keep alive per default to the istio mesh config. 
This is necessary because:
- Apps usually use http keepalive to reuse sockets. Without istio configuration, the keep alive probes would only reach istio. Istio does not forward the application keep alive probes to the hyperscaler nat-gateway.
- It applies to all outbound connections, so istio destination rules are not the correct place for the configuration.
- All (AWS, Azure, GCP) hyperscaler nat-gateways and loadbalancers have an idle-timeout (4 Minutes on [Azure](https://learn.microsoft.com/en-us/azure/load-balancer/load-balancer-tcp-reset), 350s on [AWS](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/network-load-balancers.html#connection-idle-timeout) and 10 minutes on [GCP](https://cloud.google.com/compute/docs/troubleshooting/troubleshooting-networking)).
- Hyperscaler drop connections silently (without RST or FIN frames). [Azure has a way to change the configuration](https://learn.microsoft.com/en-us/azure/load-balancer/load-balancer-tcp-reset), the others do not.
- Without this configuration, application http clients will get connection reset when reusing idle sockets besides having an application keep-alive.

The timings were choosen with a sane default which should match all hyperscalers. Most technologies would even be more frequent, like nodejs for example with a [1s default](https://nodejs.org/api/http.html#new-agentoptions))

### DIFF
--- a/manifests/istio-operator-template-light.yaml
+++ b/manifests/istio-operator-template-light.yaml
@@ -125,6 +125,10 @@ spec:
             Any modifications are discarded and the resource is reverted to the original state.
   hub: europe-docker.pkg.dev/kyma-project/prod/external/istio
   meshConfig:
+    tcpKeepalive:
+      interval: 1s
+      probes: 9
+      time: 30s
     defaultConfig:
       holdApplicationUntilProxyStarts: true
       proxyMetadata:
--- a/manifests/istio-operator-template.yaml
+++ b/manifests/istio-operator-template.yaml
@@ -138,6 +138,10 @@ spec:
             Any modifications are discarded and the resource is reverted to the original state.
   hub: europe-docker.pkg.dev/kyma-project/prod/external/istio
   meshConfig:
+    tcpKeepalive:
+      interval: 1s
+      probes: 9
+      time: 30s
     defaultConfig:
       holdApplicationUntilProxyStarts: true
       proxyMetadata:


triffer:
## Scenarios
There are two scenarios in Istio where TCP keepalives are leveraged.

**Ingress Gateway to Service Instances**: When traffic enters the mesh through the Istio Ingress Gateway, it establishes TCP connections to backend service instances.
**Service-to-Service Communication**: Within the mesh, services interact with each other. Istio also enables TCP keepalives for these service-to-service connections.

## Configure TCP keepalive in Istio
There are various mechanisms to configure the TCP keepalive.

**Global Mesh Options**: Istio provides a set of global mesh options that govern TCP keepalive behavior for all connections within the mesh.
**Destination Rule**: For specific services.
**Envoy Filters**: Filters that can intercept and modify TCP keepalive behavior can be added.

In this PR the Global Mesh Options mechanism is proposed. This means, that this is the default configurations for the whole mesh. The following lists the benefits and drawbacks of using this default.

### Benefits of default TCP Keepalives in Istio
- Proactively detect and address potential connection issues, preventing unexpected connection drops
- By proactively closing idle connections, the unnecessary resource consumption can be reduced

### Drawbacks of default TCP Keepalives in Istio
- Defaults might not be a fit for all applications and a Destination Rule must be used to overwrite the defaults.
- Default TCP keepalive settings will increase network overhead by sending keep-alive packets
- keep-alive packets can introduce additional latency into the connection, which can affect the overall throughput of the application

Additionally, we should consider exposing this configuration as part of Istio CR to have the possibility to change the defaults.

## TCP Keepalive configuration values
Some of the proposed values seem pretty low given the Istio defaults, especially the interval of 1s.
The default values used by Istio are based on the OS and for Linux they are

> time: 2h
> interval: 75s
> probes: 9

If we want to have a general configuration, we need to consider the most restricting hyperscaler LB (4 minutes idle timeout on Azure). The following configuration might be better in terms of network overhead and still respecting the limitations by the hyperscaler LBs.

> time: 200s
> interval: 75s
> probes: 9

If specific requirements exist for a particular application or scenario, such as closing connections sooner than the default 75 seconds with 9 probes, Destination Rules can be employed to override these default settings.

The default TCP keepalive settings of Linux might be sufficient for Istio proxies, as the most aggressive keepalive settings in the connection should determine the overall keepalive behaviour. However, there is no documentation on the behaviour of Envoy in this context, so further verification is required.

## ToDo's
- [ ] Discuss configuration in Istio CR.
- [ ] Agree on default handling and TCP keepalive values
- [ ] Add documentation about keepalive configuration and consider adding a tutorial for overriding the defaults

Rand0mF:
@triffer thanks for checking.
I agree within the mesh / service to service to communication the linux defaults would probably be fine. 
However I would strongly suggest using lower values regarding outbound (hyperscaler nat-gateway) connections. If a keepalive probe is not answered, linux will retry for `probes` times with a `interval` delay. Which would mean a total of 675s to detect a broken socket with your suggestion (and the application holding the socket handle will only know after that time that a socket is closed)

Application technologies usually use quite low values for those properties. For example [nodejs](https://nodejs.org/api/net.html#socketsetkeepaliveenable-initialdelay) uses  `TCP_KEEPCNT=10` and `TCP_KEEPINTVL=1` for every keepalive socket (which cannot be configured by the application).

When entering this "keepalive failure routine" network overhead does not really matter as it's an exceptional occurrence anyway.

Generally I would rather suggest having more network resilience and choosing lower values than considering network overhead for a TCP ACK package (which is basically zero). 

The linux defaults do not match a high dynamic cloud world and kyma should chose lower values.

barchw:
Hello @Rand0mF,
Generally, we suggest configuring TCP Keepalive with a [DestinationRule](https://istio.io/latest/docs/reference/config/networking/destination-rule/) Custom Resource, as it is not dependent on Istio installation parameters and is not applied cluster wide.
An example configuration would look like this:
```yaml
apiVersion: networking.istio.io/v1beta1
kind: DestinationRule
metadata:
  name: dr-keepalive
spec:
  host:  #outbound_host or a Service
  trafficPolicy:
    connectionPool:
      tcp:
        tcpKeepalive:
          interval: 1s
          probes: 9
          time: 30s
```
Is there a reason why this cannot be done in your use case?

Rand0mF:
@barchw as I wrote in the issue description, it applies to all workloads and all outbound connections. Destination rules are not feasible as a typical cluster would have dozens of microservices connecting to hundreds of different targets. It would require a huge amount of configuration with DesintationRules. Also there is basically no downside by applying it cluster wide. `Kyma is an opinionated set of Kubernetes-based modular building blocks` for which it would perfectly make sense to opinionate keep-alive to match the current state of cloud (hyperscaler).

barchw:
@Rand0mF In that case we would like to offer yet another solution, applying an EnvoyFilter that will provide the same configuration:
```yaml
apiVersion: networking.istio.io/v1alpha3
kind: EnvoyFilter
metadata:
  name: tcp-keepalive
  namespace: istio-system
spec:
  configPatches:
  - applyTo: CLUSTER
    patch:
      operation: MERGE
      value:
        upstream_connection_options:
          tcp_keepalive:
            keepalive_probes: 9
            keepalive_time: 30 #[s]
            keepalive_interval: 1 #[s]
```
Applying it to `Istio-system` ensures that it applies to all sidecars and ingress-gateways.

Rand0mF:
@barchw thanks I will try that. Still I'm wondering why you are so hesitant to add it to the default mesh config. From my point of view I can ensure there will be many issues for all kyma projects not having this configuration per default and running cloud native applications. Debugging such issues is a pain and the root cause of random connection reset errors caused by this missing configuration will probably not be found by many teams.

triffer:
We are closing this PR as we do not want to specify a default TCP keepalive setting for every mesh. If there are more requests to configure this, we will also consider adding this configuration to Istio CR.

Issue #527, opened 2023-12-20
Move prow related pipelines to GH Actions


strekm:
**Description**

As a part of https://github.com/kyma-project/test-infra/issues/9491 move Istio related prow pipelines to GH Actions. Migrate all istio jobs except operator build jobs.

ACs:
- [x] operator build job is run in prow
- [x] all istio jobs are run with GH Actions
- [x] docu updated (permalinks - let's do it in the followup)
- [x] all jobs conf stays as it was in prow
- [x] all jobs are running respectfully
- [x] all jobs are green

**Reasons**

Moving out of prow

**DoD:**
~~- [ ] Create a follow-up issue.~~
~~- [ ] Provide unit tests.~~
~~- [ ] Provide integration tests.~~
- [x] Provide documentation.
- [x] Test on a production-like environment.
~~- [ ] Verify if the solution works for both open-source Kyma and SAP BTP, Kyma runtime.~~
- [x] Check the outcome of all related pipelines.
~~- [ ] Verify resource limits.~~
~~- [ ] As a PR reviewer, verify code coverage and evaluate if it is acceptable.~~
~~- [ ] Create a release.~~
~~- [ ] Add release notes and What's New notes for Kyma customers.~~

**Attachments**

part of: https://github.com/kyma-project/test-infra/issues/9491

**PR**

- https://github.com/kyma-project/istio/pull/584
- https://github.com/kyma-project/test-infra/pull/9778

strekm:
Non AWS related pipelines can be moved

PR #584, opened 2024-01-25
Move Prow to GHA

barchw:
**Description**

- Move Goat jobs to GHA

**Related issues**
- #527


kolodziejczak (changes requested):
We are missing AWS specific integration test on main and PR to release branches

barchw:
> We are missing AWS specific integration test on main and PR to release branches

@kolodziejczak I would say we need to re-discuss that topic again
If anything we could still do it in follow up to this one, so that's how I would do it if we agree on running on PRs as well

hack/ci/gardener-integration-aws-specific.sh
```diff
@@ -0,0 +1,24 @@
...
+      -n "garden-${GARDENER_PROJECT_NAME}"
+
+    exit
+}
+
+./hack/ci/provision-gardener.sh
+# Cleanup on exit, be it successful or on fail
+trap cleanup EXIT INT
+
+make install deploy aws-integration-test
```
kolodziejczak, line 24: Make target `aws-integration-test` already contains install and deploy embedded

hack/ci/gardener-integration-aws-specific.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
```
kolodziejczak, line 1: We have already a script for that in tests/integration/scripts. I assume this is a way of duplicating scripts so prow won't fail after merging?

.github/workflows/main-integration.yaml
```diff
@@ -41,11 +41,118 @@ jobs:
...
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: ./.github/actions/integration-test
+        with:
+          operator-image-name: "europe-docker.pkg.dev/kyma-project/prod/istio-manager:${{github.sha}}"
+          servers-memory: "16g"
+          agents: 2
+
+  istio-integration-gcp:
```
kolodziejczak, line 58: we didn't have this job on post main, but in my opinion more jobs is ok

.github/workflows/main-integration.yaml
```diff
@@ -41,11 +41,118 @@ jobs:
...
+        with:
+          fetch-depth: 0
+      - run: ./tests/integration/scripts/gardener-kubeconfig.sh
+        shell: bash
+        env:
+          GARDENER_TOKEN: ${{ secrets.GARDENER_TOKEN }}
+      - uses: actions/setup-go@v4
+        with:
+          go-version: "1.21"
+      - run: make IMG="europe-docker.pkg.dev/kyma-project/prod/istio-manager:${{github.sha}}" gardener-istio-integration-test
```
kolodziejczak, line 135: This target also wasn't running on post main, but also in my opinion ok to have it here

.github/workflows/pull-integration-rel.yaml
```diff
@@ -0,0 +1,133 @@
...
+          GARDENER_CLUSTER_VERSION: "1.27"
+          GARDENER_REGION: "eu-west-1"
+          GARDENER_ZONES: "eu-west-1b,eu-west-1c,eu-west-1a"
+          MACHINE_TYPE: "m5.xlarge"
+          DISK_SIZE: 50
+          DISK_TYPE: "gp2"
+          SCALER_MAX: 3
+          SCALER_MIN: 1
+
+  istio-integration-aws:
```
kolodziejczak, line 103: job with this target was running on GCP in prow

kolodziejczak (changes requested):

kolodziejczak (approved):

Issue #623, opened 2024-02-14
External authorizer implementation


strekm:
**Description**

Based on [API proposal](https://github.com/kyma-project/istio/issues/588) enhance Istio CR. User should be able to configure http external authorizer in Istio module.  User provided input should be validated. Configuration options should be documented. User should be able to configure ext authz with kubectl and Kyma dashboard.

ACs:
- [ ] configuration of authorizers is possible
- [ ] user input is validate
- [ ] provided configuration is effective
- [ ] Istio CR UI is enhanced
- [ ] Istio tutorial written
- [ ] e2e test implemented

**Reasons**

User is able to configure Istio module with own authorizer.

**DoD:**
- [ ] Provide unit and integration tests.
- [ ] Provide documentation.
- [ ] Verify if the solution works for both open-source Kyma and SAP BTP, Kyma runtime.
- [ ] If you changed the resource limits, explain why it was needed.
- [ ] Verify that your contributions don't decrease code coverage. If they do, explain why this is the case.
- [ ] Add release notes.

**Attachments**
part of: https://github.com/kyma-project/istio/issues/549

PR #639, opened 2024-02-20
feat: Allow configuration of ExternalAuthorizer in IstioCR

barchw:
**Description**

- Allow configuration of External Authoriser in Istio CR

**Related issues**

- https://github.com/kyma-project/istio/issues/623


internal/reconciliations/external_authorizer/external_authorizer_test.go
```diff
@@ -0,0 +1,240 @@
...
+						Service: expectedService,
+						Port:    expectedPort,
+					},
+				},
+			},
+		},
+	}
+
+	Context("Duplicated authorization name", func() {
+		It("should cause a warning state", func() {
```
triffer, line 64: We are not validating that the CR is in warning state, but only that an error is returned. So we might want to adapt the test description.

internal/reconciliations/external_authorizer/external_authorizer_test.go
```diff
@@ -0,0 +1,240 @@
...
+
+	Context("ServiceEntry creation", func() {
+
+		It("should succeed creating ServiceEntry", func() {
+			//given
+			client := createFakeClient()
+			reconciler := external_authorizer.NewReconciler(client)
+
+			//when
+			err := reconciler.Reconcile(context.TODO(), istioCR)
```
triffer, line 111: I remember, that it was discussed to use context.Background instead of TODO. Is this still valid?

internal/reconciliations/external_authorizer/external_authorizer_test.go
```diff
@@ -0,0 +1,240 @@
...
+
+			// then
+			Expect(err).To(Not(HaveOccurred()))
+
+			listErr = client.List(context.TODO(), &s)
+			Expect(listErr).To(Not(HaveOccurred()))
+			Expect(s.Items).To(HaveLen(0))
+		})
+
+		It("should not owned service entries", func() {
```
triffer, line 184: The test description might be a little bit confusing.
```suggestion
		It("should not remove service entries not owned by the module", func() {
```

internal/resources/istio_resources_finder.go
```diff
@@ -89,6 +90,10 @@ func (i *IstioResourcesFinder) FindUserCreatedIstioResources() ([]Resource, erro
 			return nil, err
 		}
 		for _, item := range u.Items {
+			val, exists := item.GetLabels()[moduleLabels.ModuleLabelKey]
```
triffer, line 93: We could also move this logic into a function (e.g. `HasModuleLabel(unstructured.Unstructured)` or `IncludesModuleLabel(map[string]string)` )in the labels package, because it might come handy in the future for other implementation as well.

triffer (commented):
The following files still use api/v1alpha1, we might want to update them as well:
- [clusterconfig_test.go](https://github.com/triffer/istio/blob/6e6254594c1f6e847ca35f6e0165b6a1f1c1b178/internal/clusterconfig/clusterconfig_test.go)
- [disclaimer_annotation_test.go](https://github.com/triffer/istio/blob/main/internal/resources/disclaimer_annotation_test.go)
- [webhooks_test.go](https://github.com/triffer/istio/blob/main/internal/webhooks/webhooks_test.go)
- [gatherer.go](https://github.com/triffer/istio/blob/main/pkg/lib/gatherer/gatherer.go)
- [gatherer_test.go](https://github.com/triffer/istio/blob/main/pkg/lib/gatherer/gatherer_test.go)
- [get_test.go](https://github.com/triffer/istio/blob/main/pkg/lib/sidecars/pods/get_test.go)

api/v1alpha2/istio_structs.go
```diff
@@ -103,3 +106,76 @@ type ResourceClaims struct {
...
+
+	// Specifies the service that implements the Envoy ext_authz HTTP authorization service.
+	// The format is "[<Namespace>/]<Hostname>".
+	// The specification of "<Namespace>"
+	// is required only when it is insufficient to unambiguously resolve a service in the service registry.
+	// The "<Hostname>" is a fully qualified host name of a service defined by the Kubernetes service or ServiceEntry.
+	// The recommended format is "[<Namespace>/]<Hostname>"
+	// Example: "my-ext-authz.foo.svc.cluster.local" or "bar/my-ext-authz".
+	// +kubebuilder:validation:Required
+	Service string `json:"service"`
```
triffer, line 123: We don't support external services outside of the mesh, because we'd need to set `location: MESH_EXTERNAL` in the ServiceEntry. Do we need to update the ADR and/or have it clearly documented anywhere?

tests/integration/features/istio/production/main-suite/istio_connection.feature
triffer, line 1: I think we accidentally removed most of the integration test suites 😬

api/v1alpha2/authorizers.go
```diff
@@ -0,0 +1,74 @@
+package v1alpha2
+
+type Authorizer struct {
+	// A unique name identifying the extension authorization provider.
```
triffer, line 4: As far as I understand we don't have this validation any longer, because it was done in the external authorizer reconciler that was removed. An alternative would be to look into using CEL and CRD validation.

tests/integration/manifests/istio_cr_ext_authz_template.yaml
triffer, line 1: To have a better cohesion, I think we should get rid off the manifests dir and just move the files to the step that is using it. Also this common manifests folder is just like a those helper packages without a specific scope, that we don't want to have.

tests/integration/steps/validation_requests.go
```diff
@@ -70,6 +71,25 @@ func ValidateResponseStatusCode(ctx context.Context, path string, expectedCode i
 	return ctx, c.Get(url, asserter)
 }
 
+func ValidateHeaderStatusCode(ctx context.Context, givenHeaderName, givenHeaderValue, path string, expectedCode int) (context.Context, error) {
```
triffer, line 74: I find the name a little bit confusing. As far as I understand we only want to validate the response code, but we also send headers with the request. The function names suggests that we are also validating headers.

tests/integration/steps/deployment.go
```diff
@@ -286,6 +326,35 @@ func CreateHttpbinApplicationWithServicePort(ctx context.Context, appName, names
 	return ctx, err
 }
 
+func CreateServiceAccount(ctx context.Context, appName, namespace string) (context.Context, error) {
```
triffer, line 329: This is not used, I think we can remove it.

api/v1alpha2/authorizers.go
triffer: I think there is no way to do this validation in CEL, so we need to add it programmatically.

barchw: I will just re-add https://github.com/kyma-project/istio/blob/6c9b9fc97166832fbf8cc031d005f02df29030f5/internal/reconciliations/external_authorizer/external_authorizer.go#L40-L48

triffer: Hmm, maybe this kind of validation would be good to be kept inside the Aapi package, to not spread validations outside of the api package. So we would have CEL validation as our go-to solution and programmatic validation inside of the api package as well. 
Does that make sense to you?

barchw: For now I created a separate package in internal, I don't think it should be accessible from outside of api-gateway. WDYT?

triffer: I agree, that's a good point. 👍

barchw: OK will add units as well

docs/user/04-00-istio-custom-resource.md
```diff
@@ -14,38 +14,61 @@ You are only allowed to use one Istio CR, which you must create in the `kyma-sys
...
+| **config**                                            | object         | Specifies the configuration for the Istio installation.                                                                                                                                                                                                                                                                                     |
+| **config.authorizers**                                | \[\]authorizer | Specifies the list of external authorizers that will be configured in Istio mesh config.                                                                                                                                                                                                                                                    |
```
nataliasitko, line 39:
```suggestion
| **config.authorizers**                                | \[\]authorizer | Specifies the list of external authorizers configured in the Istio service mesh config.                                                                                                                                                                                                                                                    |
```

docs/user/04-00-istio-custom-resource.md
```diff
@@ -14,38 +14,61 @@ You are only allowed to use one Istio CR, which you must create in the `kyma-sys
...
+| **config.numTrustedProxies**                          | integer        | Specifies the number of trusted proxies deployed in front of the Istio gateway proxy.                                                                                                                                                                                                                                                       |
+
+**Authorizer**
+
+| Parameter              | Type    | Description                                                                                                                                 |
+|------------------------|---------|---------------------------------------------------------------------------------------------------------------------------------------------|
+| **name** (required)    | string  | A unique name identifying the extension authorization provider.                                                                             |
+| **service** (required) | string  | Specifies the service that implements the Envoy ext_authz HTTP authorization service. The recommended format is "[<Namespace>/]<Hostname>". |
```
nataliasitko, line 47:
```suggestion
| **service** (required) | string  | Specifies the service that implements the Envoy `ext_authz` HTTP authorization service. The recommended format is `[<Namespace>/]<Hostname>`. |
```

docs/user/04-00-istio-custom-resource.md
```diff
@@ -14,38 +14,61 @@ You are only allowed to use one Istio CR, which you must create in the `kyma-sys
...
+| **port** (required)    | integer | Specifies the port of the external authorizer on which the authorization request is made.                                                   |
```
nataliasitko, line 48:
```suggestion
| **port** (required)    | integer | Specifies the port number of the external authorizer used to make the authorization request.                                                  |
```

docs/user/04-00-istio-custom-resource.md
```diff
@@ -14,38 +14,61 @@ You are only allowed to use one Istio CR, which you must create in the `kyma-sys
...
+| **headers**            | headers | Specifies headers to be included, added or forwarded during authorization.                                                                  |
```
nataliasitko, line 49:
```suggestion
| **headers**            | headers | Specifies headers to be included, added, or forwarded during authorization.                                                                  |
```

docs/user/04-00-istio-custom-resource.md
```diff
@@ -14,38 +14,61 @@ You are only allowed to use one Istio CR, which you must create in the `kyma-sys
...
+|--------------------------|------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| **inCheck**              | object     | Defines headers to be included or added in check authorization request.                                                                                                                                                                                                                                                                                                                                                                                                                                                                   |
```
nataliasitko, line 55:
```suggestion
| **inCheck**              | object     | Defines headers to be included or added in the authorization request check.                                                                                                                                                                                                                                                                                                                                                                                                                                                                   |
```

docs/user/04-00-istio-custom-resource.md
```diff
@@ -14,38 +14,61 @@ You are only allowed to use one Istio CR, which you must create in the `kyma-sys
...
+| **inCheck.include**      | \[\]string | List of client request headers that should be included in the authorization request sent to the authorization service.                                                                                                                                                                                                                                                                                                                                                                                                                    |
```
nataliasitko, line 56:
```suggestion
| **inCheck.include**      | \[\]string | Specifies a list of client request headers that should be included in the authorization request sent to the authorization service.                                                                                                                                                                                                                                                                                                                                                                                                                    |
```

docs/user/04-00-istio-custom-resource.md
```diff
@@ -14,38 +14,61 @@ You are only allowed to use one Istio CR, which you must create in the `kyma-sys
...
+| **inCheck.add**          | key-value  | Set of additional fixed headers that should be included in the authorization request sent to the authorization service. The Key is the header name and value is the header value. Note that client request of the same key or headers specified in `Include` will be overridden.                                                                                                                                                                                                                                                          |
```
nataliasitko, line 57:
```suggestion
| **inCheck.add**          | key-value  | Defines a set of additional fixed headers that should be included in the authorization request sent to the authorization service. The key is the header's name, and the value is the header's value. It overrides a client request of the same key or headers specified in the **include** field.                                                                                                                                                                                                                                                          |
```

docs/user/04-00-istio-custom-resource.md
```diff
@@ -14,38 +14,61 @@ You are only allowed to use one Istio CR, which you must create in the `kyma-sys
...
+| **toUpstream**           | object     | Defines headers to be forwarded to the upstream.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          |
+| **toUpstream.onAllow**   | \[\]string | List of headers from the authorization service that should be added or overridden in the original request and forwarded to the upstream when the authorization check result is allowed (HTTP code 200). If not specified, the original request will not be modified and forwarded to backend as-is. Note, any existing headers will be overridden.                                                                                                                                                                                        |
```
nataliasitko, line 59:
```suggestion
| **toUpstream.onAllow**   | \[\]string | Specifies a list of headers from the authorization service that should be added or overridden in the original request and forwarded to the upstream when the authorization check result is allowed (HTTP code `200`). If not specified, the original request will not be modified and forwarded to the backend as-is. It overrides any existing headers.                                                                                                                                                                                        | 
```

docs/user/04-00-istio-custom-resource.md
```diff
@@ -14,38 +14,61 @@ You are only allowed to use one Istio CR, which you must create in the `kyma-sys
...
+| **toDownstream**         | object     | Defines headers to be forwarded to the downstream.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        |
+| **toDownstream.onAllow** | \[\]string | List of headers from the authorization service that should be added or overridden in the original request and forwarded to the upstream when the authorization check result is allowed (HTTP code 200). If not specified, the original request will not be modified and forwarded to backend as-is. Note, any existing headers will be overridden.                                                                                                                                                                                        |
```
nataliasitko, line 61:
```suggestion
| **toDownstream.onAllow** | \[\]string | Specifies a list of headers from the authorization service that should be added or overridden in the original request and forwarded to the upstream when the authorization check result is allowed (HTTP code `200`). If not specified, the original request will not be modified and forwarded to backend as-is. It overrides any existing headers.                                                                                                                                                                                   | 
```

docs/user/04-00-istio-custom-resource.md
```diff
@@ -14,38 +14,61 @@ You are only allowed to use one Istio CR, which you must create in the `kyma-sys
...
+| **toDownstream.onAllow** | \[\]string | List of headers from the authorization service that should be added or overridden in the original request and forwarded to the upstream when the authorization check result is allowed (HTTP code 200). If not specified, the original request will not be modified and forwarded to backend as-is. Note, any existing headers will be overridden.                                                                                                                                                                                        | 
+| **toDownstream.onDeny**  | \[\]string | List of headers from the authorization service that should be forwarded to downstream when the authorization check result is not allowed (HTTP code other than 200). If not specified, all the authorization response headers, except *Authority (Host)* will be in the response to the downstream. When a header is included in this list, *Path*, *Status*, *Content-Length*, *WWWAuthenticate* and *Location* are automatically added. Note, the body from the authorization service is always included in the response to downstream. |
```
nataliasitko, line 62:
```suggestion
| **toDownstream.onDeny**  | \[\]string | List of headers from the authorization service that should be forwarded to downstream when the authorization check result is not allowed (HTTP code other than `200`). If not specified, the response to the downstream will contain all the authorization response headers, except **Authority (Host)**. When a header is included in this list, **Path**, **Status**, **Content-Length**, **WWWAuthenticate**, and *Location* are automatically added. Note that the body from the authorization service is always included in the response to downstream.  | 
```

nataliasitko (approved):

PR #656, opened 2024-03-05
Fix gardener provisioning

kolodziejczak:
**Description**

- Fix gardener provisioning


barchw (approved):

werdes72 (dismissed):

barchw (approved):
Parameter will be re-added in a follow up PR to work around issue with PR_target
LGTM

PR #700, opened 2024-04-02
bug: Node-based provider discovery

Ressetkk:
/kind bug
/area service-mesh

Add small hack that tries to determine the hyperscaler based on the first provider node.
When provider is not known, returns ErrUnknownProvider error which needs to be properly handled.

Gardener clusters are usually backed by the single hyperscaler, so all nodes in the K8s cluster contain the same prefix for providerID.

The only thing it may fail is when user adds nodes manually by hand and they have different providerIDs, which is very unlikely in gardener cluster.


internal/clusterconfig/clusterconfig.go
```diff
@@ -94,6 +95,40 @@ func EvaluateClusterConfiguration(ctx context.Context, k8sClient client.Client)
...
+	// get 1st node since all nodes usually are backed by the same provider
+	n := nodes.Items[0]
+	provider := n.Spec.ProviderID
+	switch {
+	case strings.HasPrefix(provider, "aws://"):
+		return "aws", nil
+	case strings.HasPrefix(provider, "gce://"):
+		return "gce", nil
+	case strings.HasPrefix(provider, "openstack://"):
+		return "openstack", nil
```
barchw, line 127: Hmm, generally I'm not really sure about this implementation, as it returns an error for not known providers. This way we need to specify ALL our cluster providers here for it not to return an error.
I would not expect it to error in this way, as the only information we only need here is whether it's AWS or not, and not being AWS is not an error, just a information

barchw (changes requested):
Looks good, I will verify on AWS and openstack that it works

internal/clusterconfig/clusterconfig.go
Ressetkk: These kinds of code should not be in the controller to begin with. I expect that it's up to the implementer to decide whether they want to handle an error or not. I rather wanted to have the error exposed, then handled as a return variable for better visibility in the logs. For now you can remove other cases besides AWS, though I think having it more flexible makes more sense to me.

barchw (approved):

PR #708, opened 2024-04-05
bug: Node-based provider discovery (#700)

barchw:
* bug: hack: add GetClusterProvider() function

/kind bug
/area service-mesh

Add small hack that tries to determine the hyperscaler based on the first provider node. When provider is not known, returns ErrUnknownProvider error which needs to be properly handled.

Gardener clusters are usually backed by the single hyperscaler, so all nodes in the K8s cluster contain the same prefix for providerID.

The only thing it may fail is when user adds nodes manually by hand and they have different providerIDs, which is very unlikely in gardener cluster.

* cleanup: remove HyperscalerClient code

This code proved to be buggy when other provider implements EC2-compatible API. Since the provider discovery got changed it's no longer needed.

Remove HyperscalerClient implementation and its codebase and replace with Node based discovery

**Description**

- Git cherry pick #700 

**Related issues**


kolodziejczak (approved):

docs/release-notes/1.4.2.md
```diff
@@ -0,0 +1,4 @@
+
+## Bug fixes
```
nataliasitko, line 2:
```suggestion
## Bug Fixes
```

nataliasitko (approved):

PR #738, opened 2024-04-15
chore: add gitleaks to repo

Ressetkk:
/kind chore
/area ci

Add gitleaks to repo

### DIFF
--- a/.github/workflows/pull-gitleaks.yaml
+++ b/.github/workflows/pull-gitleaks.yaml
@@ -1,0 +1,23 @@
+name: pull-gitleaks
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+
+env:
+  GITLEAKS_VERSION: 8.18.2
+
+jobs:
+  scan:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Fetch gitleaks ${{ env.GITLEAKS_VERSION }}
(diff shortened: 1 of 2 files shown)


.github/workflows/pull-gitleaks.yaml
```diff
@@ -0,0 +1,23 @@
+name: pull-gitleaks
+on:
+  pull_request:
+    types: [opened, edited, synchronize, reopened, ready_for_review]
```
triffer, line 4: I think we don't need to run the action when title or body of the PR changed.
```suggestion
    types: [opened, synchronize, reopened, ready_for_review]
```

.github/workflows/pull-gitleaks.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+jobs:
+  scan:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Fetch gitleaks ${{ env.GITLEAKS_VERSION }}
+        run: curl -Lso gitleaks.tar.gz https://github.com/gitleaks/gitleaks/releases/download/v${{ env.GITLEAKS_VERSION }}/gitleaks_${{ env.GITLEAKS_VERSION }}_linux_x64.tar.gz && tar -xvzf ./gitleaks.tar.gz
```
triffer, line 17: Is there are reason why we are not using the [official gitleaks action](https://github.com/gitleaks/gitleaks-action)? The licence is also already defined in secret `GITLEAKS_LICENSE` at the organisation level. By using the official action, we could define the version of gitleaks as latest. This way we avoid the version being outdated, which I fear can easily happen with the approach proposed in this PR.

triffer: We don't use the official action, because we want to use what is recommended by Neighbours.

triffer (approved):

Issue #721, opened 2024-04-10
Re-introduce resource tests for sidecars


barchw:
**Description**
With Istio dropping IstioOperator installed-state from cluster, we need to migrate resource tests that were relying on it to a different solution.
There is no replacement for saving the Istio installation configuration on the cluster, so most likely it is needed to look at the configuration of running pods (for example, deploy `httpbin` and check sidecar resource config)

We should also think about whether we want to re-introduce tests for `egress`, as this is not a component we install, and it would require additional setup.

**Reference**
_Originally posted by @barchw in https://github.com/kyma-project/istio/pull/676#discussion_r1544365558_

PR #742, opened 2024-04-16
Re-introduce resource tests for sidecars

triffer:
**Description**

- Add new generic step to verify container resources
- Add step for istio proxy resource verification
- Remove egress-gateway resource verification, because egress is not enabled
- Remove proxy_init resource verification, because we have cni enabled per default and therefore proxy_init container will not be used

**Related issues**

- https://github.com/kyma-project/istio/issues/721

### DIFF
--- a/tests/integration/steps/container_resources.go
+++ b/tests/integration/steps/container_resources.go
@@ -1,0 +1,96 @@
+package steps
+
+import (
+	"context"
+	"fmt"
+	"github.com/kyma-project/istio/operator/tests/integration/testcontext"
+	"github.com/pkg/errors"
+	"gopkg.in/inf.v0"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"strconv"
+	"strings"
+)
+
+type resourceStruct struct {
+	Cpu    resource.Quantity
+	Memory resource.Quantity
+}
+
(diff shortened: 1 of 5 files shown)


tests/integration/steps/container_resources.go
```diff
@@ -0,0 +1,96 @@
...
+	"strconv"
+	"strings"
+)
+
+type resourceStruct struct {
+	Cpu    resource.Quantity
+	Memory resource.Quantity
+}
+
+func DeploymentHasPodWithContainerResourcesSetToCpuAndMemory(ctx context.Context, depName, depNamespace, container, resourceType, cpu, memory string) error {
```
barchw, line 21: Function name might be a little bit disguising since it doesn't really follow the deployment by OwnerReference by deployment name, and it uses the `app: label` selector instead, which might not always be present

barchw (changes requested):

tests/integration/steps/container_resources.go
triffer: We use a similar approach when reading logs, so I thought about keeping it simple for the tests as well. But you are right, it might be confusing. Instead of iterating through OwnerReference (and adding a lot more lines), I think using the matchLabels gives the same level of confidence.

triffer:
I decided to not use the new `DeploymentHasPodWithContainerResourcesSetToCpuAndMemory` to verify istio components to keep it simple, since we don't need to check the pod.

barchw (approved):

PR #762, opened 2024-04-22
Decouple restarter

barchw:
**Description**

- ...

**Related issues**

### DIFF
--- a/pkg/lib/ingress_gateway/ingress_gateway.go
+++ b/pkg/lib/ingress_gateway/ingress_gateway.go
@@ -1,0 +1,68 @@
+package ingressgateway
+
+import (
+	"context"
+	"encoding/json"
+	operatorv1alpha2 "github.com/kyma-project/istio/operator/api/v1alpha2"
+	v1 "k8s.io/api/core/v1"
+
+	"github.com/kyma-project/istio/operator/internal/filter"
+)
+
+type IngressGatewayRestartPredicate struct {
+	istioCR *operatorv1alpha2.Istio
+}
+
+func NewIngressGatewayRestartPredicate(istioCR *operatorv1alpha2.Istio) *IngressGatewayRestartPredicate {
+	return &IngressGatewayRestartPredicate{istioCR: istioCR}
+}
+
+type appliedConfig struct {
+	operatorv1alpha2.IstioSpec
+	IstioTag string
+}
+
+func getLastAppliedConfiguration(istioCR *operatorv1alpha2.Istio) (appliedConfig, error) {
+	lastAppliedConfig := appliedConfig{}
+	if len(istioCR.Annotations) == 0 {
+		return lastAppliedConfig, nil
+	}
+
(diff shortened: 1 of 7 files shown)


pkg/lib/ingress_gateway/ingress_gateway.go
```diff
@@ -0,0 +1,67 @@
+package ingressgateway
+
+import (
+	"context"
+	"encoding/json"
+	operatorv1alpha2 "github.com/kyma-project/istio/operator/api/v1alpha2"
+	"github.com/kyma-project/istio/operator/internal/filter"
+	"github.com/kyma-project/istio/operator/internal/reconciliations/istio"
```
triffer, line 8: This lib now has a dependency on `internal/reconciliations/istio`, we should avoid this. I think we can also move all the lastAppliedConfiguration code out of the Istio package, because as far as I can see it's only used by restarters and istio_Controller.

triffer:
I think in general we should discuss how we want to proceed with the pkg and internal. I'd remove one of those and simplify the structure.

barchw:
> I think in general we should discuss how we want to proceed with pkg and internal directories. I'd remove one of those and simplify the structure.

Yes, I think we can remove the `pkg` directory now, as it was used for some time by old reconciler, but currently it does not make sense to expose the functionalities outside

triffer (dismissed):

docs/contributor/04-10-technical-design.md
```diff
@@ -120,8 +120,9 @@ This component covers the following restart triggers:
 - Restart Pods with proxy sidecar when CNI config changes.
 - Restart Pods with proxy sidecar after an Istio version update.
 - Restart Pods with proxy sidecar when proxy resources change.
-- Restart Pods if they match [Restart Predicates](#restart-predicates) that the [Istio ResourcesReconciliation component](#istio-resourcesreconciliation) specifies (for example, being up-to-date with EnvoyFilter).
+- Restart Pods if they match [Restart Predicates](#restart-predicates) that the [Istio ResourcesReconciliation component](#istio-resourcesreconciliation) specifies (for example, being up to date with proxy image version).
 
 ### IngressGatewayRestarter
 
-IngressGateway Restarter is responsible for restarting Istio Ingress Gateway. The component consumes a list of [Restart Predicates](#restart-predicates) that determine when the restart should occur.
+IngressGateway Restarter is responsible for restarting Istio Ingress Gateway. The component consumes a list of [Restart Predicates](#restart-predicates) that determine when the restart should occur. Restarter triggers the restart in case that `numTrustedProxies` configuration changed.
```
nataliasitko, line 127:
```suggestion
IngressGateway Restarter is responsible for restarting Istio Ingress Gateway. The component consumes a list of [Restart Predicates](#restart-predicates) that determine when the restart should occur. Restarter triggers the restart if there's a change in the `numTrustedProxies` configuration.
```

triffer (approved):

nataliasitko (approved):

PR #803, opened 2024-05-07
Switch to the default cni pair pods configuration to repair the pods

videlov:
**Description**

- Switch to the default cni pair pods configuration to repair the pods
https://istio.io/latest/docs/setup/additional-setup/cni/#race-condition--mitigation

**Related issues**


barchw (approved):
Doesn't seem to fail anymore, if tests become flaky again we will see that it was this that caused an issue

PR #841, opened 2024-05-24
Move Istio installation to external process

barchw:
**Description**

- Move Istio installation to external process


triffer (dismissed):
I think we should also have an ADR for this to record why we are using this external process. But we scan do it in a follow up.

werdes72 (approved):

Issue #801, opened 2024-05-07
Evaluation of the use of Istio compatibility versions and feature flags


triffer:
**Description**
With release [1.21.0 Istio introduced compatibility versions](https://istio.io/latest/news/releases/1.21.x/announcing-1.21/#easing-upgrades-with-compatibility-versions). This feature give users an additional option, allowing release versions to be decoupled from behavioral changes.

There are additional information in the [Istio documentation](https://istio.io/latest/docs/setup/additional-setup/compatibility-versions/) about compatibility versions:
> Compatibility versions are recommended to be used only when an incompatibility is found, rather than as the default. Each 
> compatibility version will only persist for a few releases, so eventually you will need to migrate to the new behavior. Currently, 
> each compatibility version is intended to remain for at least two versions, though this is subject to change.

Additionally, Istio also allows to control the [feature flags as environmental variables](https://istio.io/latest/docs/reference/commands/pilot-discovery/#envvars) of the istiod deployment.

We should evaluate how we can use these options to avoid breaking changes during rollout of Istio. 

ACs:
- [x] `compatibilityMode` introduced in Istio CR
- [x] `compatibilityMode` default setting is false
~~- [x] setting `compatibilityMode` to true sets `compatibilityVersion` to previous minor version~~
- [x] setting `compatibilityMode` to true sets list of specific env variables for previous minor version
- [x] Istio module root documentation version is extended to describe `compatibilityMode` details
- [x] Istio CR UI is extended to support `compatibilityMode` flag
- [x] Istio update GH Issue template is extended to handle management of `compatibilityMode` 
- [x] Create ADR

**Reasons**
During rollout of Istio module version 1.5.1 a [breaking change](https://istio.io/latest/news/releases/1.21.x/announcing-1.21/upgrade-notes/#externalname-support-changes) was rolled out that caused issues. We should look into options to avoid this in the future.

**DoD:**
- [x] Provide unit and ~~integration tests.~~
- [x] Provide documentation.
- [x] Verify if the solution works for both open-source Kyma and SAP BTP, Kyma runtime.
- [x] If you changed the resource limits, explain why it was needed.
- [x] If the default configuration of Istio Operator has been changed, you performed a manual upgrade test to verify that the change can be rolled out correctly.
- [x] Verify that your contributions don't decrease code coverage. If they do, explain why this is the case.
- [x] Add release notes.

**Attachments**
- https://github.com/kyma-project/istio/issues/787

[...]

PR #849, opened 2024-05-28
feat: compatibility mode

kolodziejczak:
**Description**

- Add compatibility mode

**Related issues**

- https://github.com/kyma-project/istio/issues/801


api/v1alpha2/istio_merge.go
```diff
@@ -23,6 +23,17 @@ const (
 	requestsField         = "requests"
 )
 
+// compatibility environment variables of Istio
+// they are supposed to be feature flags, to keep the compatibility with previous Istio versions in case of breaking changes
+// should be updated with every Istio bump according to the release notes
+// requires update of integration tests
+// current env comes from: Istio 1.21
+var compatibilityEnvs = map[string]string{
+	"ENABLE_EXTERNAL_NAME_ALIAS": "false",
```
triffer, line 32: As far as I know we already have this feature flag with `false` only for one cluster right now by applying the annotation. Why do we have it here as default for compatibility mode? I think we ned to exclude it from compatibility mode.

api/v1alpha2/istio_merge.go
```diff
@@ -40,13 +51,42 @@ func (i *Istio) MergeInto(op iopv1alpha1.IstioOperator) (iopv1alpha1.IstioOperat
...
+	}
+
 	return externalNameAliasAnnotationFixOp, nil
 }
 
 type meshConfigBuilder struct {
 	c *meshv1alpha1.MeshConfig
 }
 
+func setCompatibilityEnvs(_ *Istio, op iopv1alpha1.IstioOperator) iopv1alpha1.IstioOperator {
```
triffer, line 66: Why do we have a param thast we ignore `_ *Istio`?

api/v1alpha2/istio_merge.go
```diff
@@ -40,13 +51,42 @@ func (i *Istio) MergeInto(op iopv1alpha1.IstioOperator) (iopv1alpha1.IstioOperat
...
+		op.Spec.Components = &v1alpha1.IstioComponentSetSpec{}
+	}
+	if op.Spec.Components.Pilot == nil {
+		op.Spec.Components.Pilot = &v1alpha1.ComponentSpec{}
+	}
+	if op.Spec.Components.Pilot.K8S == nil {
+		op.Spec.Components.Pilot.K8S = &v1alpha1.KubernetesResourcesSpec{}
+	}
+
+	for k, v := range compatibilityEnvs {
```
triffer, line 80: We are not using the compatibilityVersion during install. This means we always need to manage a list of ENV Vars. As far as I understood we wanted to decouple as far as possible from this Istio internals and use the [compatibilityVersion](https://istio.io/latest/news/releases/1.21.x/announcing-1.21/#easing-upgrades-with-compatibility-versions) instead and additionally allow configuring env vars that are not covered by the compatibilityVersion, e.g. `ISTIO_DELTA_XDS` for version 1.22.

tests/integration/features/istio/production/main-suite/istio_configuration.feature
```diff
@@ -61,3 +61,10 @@ Feature: Configuration of Istio module
     And Log of container "ext-authz" in deployment "ext-authz" in namespace "default" contains "X-Add-In-Check:[value] X-Ext-Authz:[allow]"
     And Request with header "x-ext-authz" with value "deny" to path "/headers" should have response code "403"
     And Log of container "ext-authz" in deployment "ext-authz" in namespace "default" contains "X-Add-In-Check:[value] X-Ext-Authz:[deny]"
+  Scenario: Compatibility mode is on
```
triffer, line 64: I don't like that we have tests that we need to update on every Istio version update, but there might be no good way to observe a different behaviour for compatibilityVersion. I'd also have this as a separate feature and not add it to the Istio configuration. Also some of those compatibilityVersion changes might not be observable by checking ENV Variables, what do we do then? An example for this is the [default tracing changes in 1.22](https://istio.io/latest/news/releases/1.22.x/announcing-1.22/upgrade-notes/#default-tracing-to-zipkinistio-systemsvc-removed). In this case this test would be completely invalid if only such changes are controlled by the compatibilityVersion.

triffer (changes requested):

triffer:
We are also missing a general documentation about what compatibility mode is and how it should be used.

api/v1alpha2/istio_merge.go
kolodziejczak: But then, if a new cluster want to go into the compatibility mode, it has to also apply the annotation to fully achieve that, which is counter intuitive. Current behaviour let you use annotation to apply only this one env variable with false value, but if you decide to to turn compatibility mode on, then it's applied as well.

api/v1alpha2/istio_merge.go
kolodziejczak: Right, I'll adjust to rely mainly on the compatibilityVersion

tests/integration/features/istio/production/main-suite/istio_configuration.feature
triffer: We might want to update this tests to verify that those env vars are not sent before we switch the compatibility mode one and then verify that they are set. This will require to update this test with every Istio release. I think there is no other way to have the verification that the flag we provide to Istio works as expected. 
Also there might be the case where the compatibility version doesn't use any env vars, so this test might not be applicable.

kolodziejczak: as we discussed, let's go with this approach

tests/integration/steps/deployment.go
```diff
@@ -329,6 +330,55 @@ func CreateServiceWithPort(ctx context.Context, appName, namespace string, port,
...
+				if e.Name == envVarName && e.Value == envVarVal {
+					return nil
+				}
+			}
+		}
+		return fmt.Errorf("expected env variable %s on Deployment %s with value %s", envVarName, deploymentName, envVarVal)
+	}, testcontext.GetRetryOpts()...)
+}
+
+func VerifyLackEnvVariableOnDeployment(ctx context.Context, envVarName, deploymentName, namespace string) error {
```
triffer, line 357: Maybe a more expressive name is easier to understand `VerifyEnvVariableIsNotOnDeployment`?

tests/integration/features/istio/production/main-suite/istio_compatibility_mode.feature
```diff
@@ -0,0 +1,24 @@
...
+    Given "Istio CR" is not present on cluster
+    And Evaluated cluster size is "Production"
+    And Istio CRD is installed
+    And Namespace "istio-system" is "not present"
+    And "Deployment" "istio-controller-manager" in namespace "kyma-system" is ready
+
+  Scenario: Compatibility mode is set on
+    Given Istio CR "istio-sample" from "istio_cr_template" is applied in namespace "kyma-system"
+    And Istio CR "istio-sample" in namespace "kyma-system" has status "Ready"
+    # ENABLE_EXTERNAL_NAME_ALIAS is set to true by the custom annotation handler, because of that we need to verify if it's not overwritten
```
triffer, line 13: This is confusing as we are verifying here that it's true and stating we need to verify that it's not overwritten and later we check if it has value false.

kolodziejczak: It's the only compatibility variable affected by the custom annotation impl. We could also drop this step, but I decided to include it so we have a full coverage of the behaviour. Also I can improve the comment, to more precisely describe what's going on. What's your proposal?

tests/integration/steps/deployment.go
kolodziejczak: we can do that

internal/reconciliations/istio/external_installation/external_installation.go
```diff
@@ -0,0 +1,76 @@
+package external_installation
```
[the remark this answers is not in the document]
kolodziejczak: I feel like it's cleaner to separate it with a new package. Do you see some cons that would favour staying within the same package?

internal/reconciliations/istio/external_installation/external_installation.go
```diff
@@ -0,0 +1,76 @@
...
+	major := sp[0]
+
+	tmp, err := strconv.Atoi(sp[1])
+	if err != nil {
+		return "", err
+	}
+
+	minor := strconv.Itoa(tmp - 1)
+
+	compatibilityParam := flag + major + semVerSep + minor
```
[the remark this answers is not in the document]
kolodziejczak: definitely!

api/v1alpha2/merge_test.go
```diff
@@ -539,6 +538,102 @@ var _ = Describe("Merge", func() {
...
+				},
+			}
+
+			// when
+			out, err := istioCR.MergeInto(iop)
+
+			//then
+			Expect(err).ShouldNot(HaveOccurred())
+
+			vc := 0
```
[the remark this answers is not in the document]
kolodziejczak: it's variable counter, we can do more expressive

internal/reconciliations/istio/external_installation/external_installation_test.go
```diff
@@ -0,0 +1,53 @@
...
+			Expect(err).ToNot(HaveOccurred())
+			Expect(ei.Args[2]).To(Equal(""))
+		})
+
+		It("should fail if istio version is not following the semver", func() {
+			iopPath := "/some/path"
+			compatibilityMode := true
+			istioVersion := "1.21"
+			_, err := external_installation.NewExternalInstaller(iopPath, istioVersion, compatibilityMode)
+			Expect(err).To(HaveOccurred())
```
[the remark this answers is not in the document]
kolodziejczak: Refactored to custom error

internal/reconciliations/istio/external_installation/external_installation.go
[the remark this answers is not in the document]
triffer: A package with just one file seems like not a good idea and the external installer has a high cohesion with code in existing istio package , that's why I would keep it simple for now and keep it there. I also don't see any usecase to use the external isntaller outside of the istio package.

tests/integration/features/istio/production/main-suite/istio_compatibility_mode.feature
triffer: You should keep the step, but the comment is confusing for me.

kolodziejczak: Then I'll improve it

internal/reconciliations/istio/external_installation/external_installation.go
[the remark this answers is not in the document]
kolodziejczak: Moved

triffer (approved):

docs/user/00-10-overview-istio-controller.md
```diff
@@ -92,4 +92,8 @@ Run this command to get all resources created by the Istio module:
 
 ```bash
 kubectl get all|<resources-kind> -A -l kyma-project.io/module=istio
-```
\ No newline at end of file
+```
+
+## Compatibility Mode
+
+You can set Istio Module to work in the compatibility mode by setting IstioCR's `spec.compatibilityMode` field. It let's users to keep possibility to mitigate breaking changes during the Istio bump. Istio Module supports compatibilityVersion one minor back to the current Istio version (eg. Istio module with Istio 1.21.0 applies compatibilityVersion to 1.20). See [Istio Documentation on compatibilityVersion](https://istio.io/latest/docs/setup/additional-setup/compatibility-versions/).
```
nataliasitko, line 99:
```suggestion
To enable compatibility mode in the Istio module, you can set the **  spec.compatibilityMode** field in the Istio CR. This allows you to mitigate breaking changes when a new release introduces an Istio upgrade. The Istio module supports compatibility with the previous minor version of Istio. For example, the Istio module with Istio 1.21.0 applies a compatibility version of Istio 1.20. For more information, see [Compatibility Versions](https://istio.io/latest/docs/setup/additional-setup/compatibility-versions/).
```

docs/user/04-00-istio-custom-resource.md
```diff
@@ -18,6 +18,7 @@ This table lists all the possible parameters of the given resource together with
...
 |-------------------------------------------------------------|----------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| **compatibilityMode**                                       | bool           | Sets Istio in a compatibility mode to remain compatible with one minor version behind.                                                                                                                                                                                                                                                                                    |
```
nataliasitko, line 21:
```suggestion
| **compatibilityMode**                                       | bool           | Enables compatibility mode in Istio. When the value is `true`, the Istio module uses a compatibility version set to the previous minor version of Istio.                                                                                                                                                                                                                                  |
```

.github/ISSUE_TEMPLATE/istio_update.md
```diff
@@ -20,6 +20,8 @@ ACs:
 - [ ] Tests and documentation updated if needed.
 - [ ] Verify that sidecars are in sync with Control Plane.
 - [ ] Istio and Envoy Version updated in the [`README.md`](https://github.com/kyma-project/istio) and [`/docs/user/README.md`](https://github.com/kyma-project/istio) files. You can use the scripts `scripts/get_module_istio_version.sh` and `scripts/get_module_envoy_version.sh` to extract the versions.
+- [ ] Check compatibilityVersion, for one minor back, and adjust integration tests, to cover environment variable contained there. You can find it [here](https://github.com/istio/istio/tree/master/manifests/helm-profiles).
+- [ ] Check compatibility variables not covered by compatibilityVersion parameter for the new version, and update them if neccessary.
```
nataliasitko, line 24:
```suggestion
- [ ] Check compatibility variables not covered by **compatibilityVersion** parameter for the new version. Update them if necessary. 
```

.github/ISSUE_TEMPLATE/istio_update.md
```diff
@@ -20,6 +20,8 @@ ACs:
 - [ ] Tests and documentation updated if needed.
 - [ ] Verify that sidecars are in sync with Control Plane.
 - [ ] Istio and Envoy Version updated in the [`README.md`](https://github.com/kyma-project/istio) and [`/docs/user/README.md`](https://github.com/kyma-project/istio) files. You can use the scripts `scripts/get_module_istio_version.sh` and `scripts/get_module_envoy_version.sh` to extract the versions.
+- [ ] Check compatibilityVersion, for one minor back, and adjust integration tests, to cover environment variable contained there. You can find it [here](https://github.com/istio/istio/tree/master/manifests/helm-profiles).
```
nataliasitko, line 23:
```suggestion
- [ ] Check **compatibilityVersion** of the previous minor version and adjust integration tests to cover the environment variable contained there. You can find it in the [`helm-profiles`](https://github.com/istio/istio/tree/master/manifests/helm-profiles) directory.
```

docs/user/04-00-istio-custom-resource.md
```diff
@@ -82,6 +83,9 @@ This table lists all the possible parameters of the given resource together with
 
 To retain the behavior of `EXTERNAL_NAME` that was present in versions of Istio prior to 1.21, you can configure the `ENABLE_EXTERNAL_NAME_ALIAS` environment variable in the Istio pilot. To do this, add the following annotation to the Istio CR:
 
+!> [!WARNING]
+> Annotation does not have effect when compatibility mode is on. The `ENABLE_EXTERNAL_NAME_ALIAS` environment variable in this mode is set to false.
+
```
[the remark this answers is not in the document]
kolodziejczak: @nataliasitko I updated suggestion just a little bit, please check if it's ok

triffer:
The current compatibilityVersion will not override flags defined in the provided IstioOperator.yaml. Unfortunately, the compatibilityVersion handling implemented in Istio will add new flags that do not exist in the IstioOperator. With the compatibilityVersion 1.21, the [zipkin address is set](https://github.com/istio/istio/blob/24e32e1895a3451b386cb767f27b223dc7323e4e/manifests/helm-profiles/compatibility-version-1.20.yaml#L20). This was removed from the IstioOperator in our module and enabling the compatibilityVersion will bring it back. Due to this, it seems like we cannot use the compatibilityVersion, but need to control this programatically when creating the IstioOperator in the module.

api/v1alpha2/compatibility_mode.go
```diff
@@ -0,0 +1,40 @@
...
+		op.Spec.Components = &v1alpha1.IstioComponentSetSpec{}
+	}
+	if op.Spec.Components.Pilot == nil {
+		op.Spec.Components.Pilot = &v1alpha1.ComponentSpec{}
+	}
+	if op.Spec.Components.Pilot.K8S == nil {
+		op.Spec.Components.Pilot.K8S = &v1alpha1.KubernetesResourcesSpec{}
+	}
+
+	for k, v := range pilotCompatibilityEnvVars {
```
triffer, line 32: When using Istio compatibility version, there is the behaviour, that configuration existing in our Istio Operator are not overwritten. I think we don't need this general behaviour and it would be a decision when implementing the compatibility mode for the specific version. Do you agree?

api/v1alpha2/compatibility_mode_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	operatorv1alpha1 "istio.io/api/operator/v1alpha1"
+	iopv1alpha1 "istio.io/istio/operator/pkg/apis/istio/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = Describe("Merge", func() {
```
triffer, line 11: Maybe we can make this more specific by having this Comaptibility Mode on the top level and scoping the level below to Istio Pilot Env Vars. This way it might be easier to distinguish, when we start implementing flags on a different level, e.g. we could have a new Context "Proxymetadata".
```suggestion
var _ = Describe("Compatibility Mode", func() {
    Context("Pilot Env Variables", func() {
```

docs/contributor/adr/0007-compatibilityMode.md
```diff
@@ -0,0 +1,52 @@
+# Support for **compatibilityMode** Configuration
+
+## Status
+
+Accepted by the Istio module team
```
triffer, line 5: Status should only be Accepted, I just saw that it's also different for the [Support for externalTrafficPolicy Configuration ADR](https://github.com/kyma-project/istio/blob/main/docs/contributor/adr/0006-support-for-externalTrafficPolicy-configuration.md), maybe you can also adapt that ADR.
```suggestion
Accepted
```

docs/contributor/adr/0007-compatibilityMode.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Context
+
+With Istio's introduction of a [compatibilityVersion](https://istio.io/latest/news/releases/1.21.x/announcing-1.21/#easing-upgrades-with-compatibility-versions) and environment variables allowing to mitigate breaking changes during Istio upgrade, Istio module team decided to introduce compatibility mode to allow users leverage possibility of remaining compatible with previous Istio versions.
+
+## Considerations
```
triffer, line 11: Consideration is not part of our ADR template 🤔 This is just part of the decision, but you might need to write it different. I would not write it as a question, but mention what were the options, what was the decision and why the decision was made.

docs/contributor/adr/0007-compatibilityMode.md
```diff
@@ -0,0 +1,52 @@
...
+
+Should `compatibilityVersion` be used, exposed by Istio repository with an additional possibility to add environment variables, or a custom implementation allowing Istio team to control the scope of the compatibility settings, either in the environment variables or other parts of the IstioOperator config?
+
+## Decision
+
+To avoid a situation, in which compatibilityVersion introduces changes that have breaking, or unwanted impact on the Istio configuration, Istio team decide not to use `compatibilityVersion`, and to create custom implementatation.
```
triffer, line 17: I'm missing that this decision was mainly done, because compatibility version would again introduce configuration that was on purpose already removed from the Istio operator in module before.

triffer, line 17: What does custom implementation mean in this case? Maybe be more specific and describe it similar to "but introduce an implementation that will add the flags controlled by the compatibilityVersion to the IstioOperator configuration created in the controller programatically." After this we should mention the result, e.g. "That allows a more fine-grained control of the flags applied for the compatibility mode"

docs/contributor/adr/0007-compatibilityMode.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Consequences
+
+With every Istio major or minor version bump, Istio module team has to manually take care of a proper content of the compatibility settings.
+
+### Sample configuration:
```
triffer, line 23: I would not add this examples since they provide not much to the original issue we were facing, with overriding configuration.

.github/ISSUE_TEMPLATE/istio_update.md
```diff
@@ -20,6 +20,8 @@ ACs:
 - [ ] Tests and documentation updated if needed.
 - [ ] Verify that sidecars are in sync with Control Plane.
 - [ ] Istio and Envoy Version updated in the [`README.md`](https://github.com/kyma-project/istio) and [`/docs/user/README.md`](https://github.com/kyma-project/istio) files. You can use the scripts `scripts/get_module_istio_version.sh` and `scripts/get_module_envoy_version.sh` to extract the versions.
+- [ ] Check **compatibilityVersion** of the previous minor version . You can find it in the [`helm-profiles`](https://github.com/istio/istio/tree/master/manifests/helm-profiles) directory. Evaluate content, and adjust compatibility mode implementation. You can find it in the v1alpha2 package(api/v1alpha2/).
+- [ ] Check compatibility variables not covered by **compatibilityVersion** parameter for the new version in the Istio release notes. Update them if necessary.
```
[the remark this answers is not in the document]
kolodziejczak: Verify if there are any compatibility related env variables in the release notes, that are not listed in the compatibilityVersion yaml. This way or another we look into that, so I'd thought about it as a reminder to pay attention to such cases.

api/v1alpha2/compatibility_mode_test.go
kolodziejczak: Definitely

api/v1alpha2/compatibility_mode.go
kolodziejczak: I do

docs/user/00-10-overview-istio-controller.md
```diff
@@ -92,4 +92,8 @@ Run this command to get all resources created by the Istio module:
 
 ```bash
 kubectl get all|<resources-kind> -A -l kyma-project.io/module=istio
-```
\ No newline at end of file
+```
+
+## Compatibility Mode
+
+To enable compatibility mode in the Istio module, you can set the **spec.compatibilityMode** field in the Istio CR. This allows you to mitigate breaking changes when a new release introduces an Istio upgrade. The Istio module supports compatibility with the previous minor version of Istio. For example, the Istio module with Istio 1.21.0 applies a compatibility version of Istio 1.20. For more information, see [Compatibility Versions](https://istio.io/latest/docs/setup/additional-setup/compatibility-versions/).
```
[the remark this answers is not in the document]
kolodziejczak: would you see them listed per Istio version or per Istio Module version?

docs/contributor/adr/0007-compatibilityMode.md
```diff
@@ -0,0 +1,52 @@
+# Support for **compatibilityMode** Configuration
+
+## Status
+
+Accepted by the Istio module team
+
+## Context
+
+With Istio's introduction of a [compatibilityVersion](https://istio.io/latest/news/releases/1.21.x/announcing-1.21/#easing-upgrades-with-compatibility-versions) and environment variables allowing to mitigate breaking changes during Istio upgrade, Istio module team decided to introduce compatibility mode to allow users leverage possibility of remaining compatible with previous Istio versions.
```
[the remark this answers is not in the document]
kolodziejczak: To me it does not fit well into the Context part, I'd add it but in the decision part.

docs/contributor/adr/0007-compatibilityMode.md
kolodziejczak: Then I'd remove this part, because we have other part that explains decision

.github/ISSUE_TEMPLATE/istio_update.md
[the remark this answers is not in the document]
triffer: Maybe we can merge it with the other AC you added, because they are both about compatibility version.

docs/user/00-10-overview-istio-controller.md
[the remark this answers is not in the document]
triffer: I think we should just have the current flags mentioned without a reference to the module version, because we will probably have them mentioned as well in the release notes or at least in the What's New, so users can decide rather they want the compatibility mode or not.

docs/contributor/adr/0007-compatibilityMode.md
```diff
@@ -6,47 +6,13 @@ Accepted
...
+To avoid a situation, in which compatibilityVersion introduces changes that have a breaking, or an unwanted impact on the Istio configuration, Istio team decided to not to use the `compatibilityVersion`, and to create a custom implementatation. The custom implementation in the Istio Module controller applies a subset of the Istio compatibilityVersion flags that are examined and extracted by the Istio team, and that fits with the configuration provided by the Istio Module. This decision was made, because Istio compatibilityVersion 1.21 would introduce configuration that was on purpose already removed from the IstioOperator before, and there is no guarantee that the following compatibityVersion of the Istio in the future, would not include breaking changes for the Istio Module opinionated configuration. Since compatibility support is meant for one minor version back, the compatibility mode from the user perspective has been decided to be a boolean flag on the IstioCR.
```
triffer, line 13: I like that 👍

triffer (approved):

triffer:
Before considering that as done and merging we should discuss if we are good with the flow for the user in the managed scenario. Since the current implementation is only a boolean flag and the user in the managed scenario cannot define when the new Istio version is rolled out, the user must enable/disable the compatibilityMode before or after the Istio rollout.
The question how the user will use this flag. If we make the assumption that the main use case is, that the user will see breaking changes in the release notes, enable compatibility mode to not have issues once the new version is released, we also need to consider that this will already enable compatibility mode for the Istio version currently running on a cluster and therefore might lead to issues. 
The compatibility mode in the current mode mainly helps users by having more time to adapt the changes by having fast channel with compatibility mode off and regular channel with compatibility mode on.
If we agree on that as our main scenario, then we might want to add it to the documentation to be more specific about this.

triffer:
Before we consider this done and merge, let's discuss if we agree with the flow for the user in the managed scenario. Since the current implementation is only a boolean flag and the user in the managed scenario cannot define when the new Istio version will be rolled out, the user needs to enable/disable the compatibilityMode before or after the rollout of Istio.
The question is how the user will use this flag. If we assume that the main use case is that the user will see the changes in the release notes and enable compatibility mode to not have any issues once the new version is released, we also have to consider that this will already enable compatibility mode for the Istio version currently running on a cluster and therefore may cause issues as it's introducing changes for the older Istio version.
 
Compatibility mode in the current implementation helps users mainly by giving them more time to adapt to the changes by having a fast channel with compatibility mode off and a regular channel with compatibility mode on.
If we agree that this is our main scenario, we should include it in the documentation to explain this in more detail.

docs/contributor/adr/0007-compatibilityMode.md
kolodziejczak: glad to hear that!

docs/user/00-10-overview-istio-controller.md
```diff
@@ -94,4 +94,11 @@ Run this command to get all resources created by the Istio module:
...
-```
\ No newline at end of file
+```
+
+## Compatibility Mode
+
+To enable compatibility mode in the Istio module, you can set the **spec.compatibilityMode** field in the Istio CR. This allows you to mitigate breaking changes when a new release introduces an Istio upgrade. The Istio module applies an opinionated subset of Istio compatibilityVersion, and supports compatibility with the previous minor version of Istio. For example, the Istio module with Istio 1.21.0 applies a compatibility version of Istio 1.20. For more information, see [Compatibility Versions](https://istio.io/latest/docs/setup/additional-setup/compatibility-versions/).
+
+> [!WARNING]
+> In case when previous version behaviour is not compatible with the current mesh setup, turning on the compatibility mode might break some of the configs until the release of the new version is done.
```
nataliasitko, line 104:
```suggestion
> If you want to use the compatibility mode to retain the current version of Istio when a new version of the Istio module is released, the compatibilityVersion will be first set to a minor version lower than the one you are currently using. If this lower version is not compatible with your current mesh setup, some configurations may be broken until the new release of the Istio module is rolled out.
```

nataliasitko (approved):

PR #852, opened 2024-05-29
Remove deepcopy from sidecar restart

barchw:
**Description**

- Remove deepcopy from sidecar restart to decrease memory usage

**Related issues**

### DIFF
--- a/pkg/lib/sidecars/pods/get.go
+++ b/pkg/lib/sidecars/pods/get.go
@@ -51,14 +51,11 @@ func getAllRunningPods(ctx context.Context, c client.Client) (*v1.PodList, error
 	return podList, nil
 }
 
-func GetPodsToRestart(ctx context.Context, c client.Client, expectedImage SidecarImage, expectedResources v1.ResourceRequirements, predicates []filter.SidecarProxyPredicate, logger *logr.Logger) (outputPodsList v1.PodList, err error) {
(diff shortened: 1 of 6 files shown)


pkg/lib/sidecars/pods/get.go
```diff
@@ -51,32 +51,32 @@ func getAllRunningPods(ctx context.Context, c client.Client) (*v1.PodList, error
...
 		if err != nil {
-			return v1.PodList{}, err
+			return &v1.PodList{}, err
 		}
 
+		outputPodsList = &v1.PodList{}
 		for _, pod := range podList.Items {
 			if evaluator.RequiresProxyRestart(pod) {
-				outputPodsList.Items = append(outputPodsList.Items, *pod.DeepCopy())
+				outputPodsList.Items = append(outputPodsList.Items, pod)
```
triffer, line 72: Maybe a micro-optimization since the pod objects are not that big, but I think we will create multiple backing arrays, since the capacity of the slices will increase multiple times. That will lead to multiple arrays being kept in memory until the outputPodList isn't referenced any longer and can be garbage collected.
We could initialize the `outputPodsList.Items` with a capacity derived from podList.Items length.

main.go
```diff
@@ -107,6 +109,16 @@ func main() {
 		HealthProbeBindAddress: flagVar.probeAddr,
 		LeaderElection:         flagVar.enableLeaderElection,
 		LeaderElectionID:       "76223278.kyma-project.io",
+		Client: client.Options{
+			Cache: &client.CacheOptions{
+				DisableFor: []client.Object{
```
triffer, line 114: Should we add a comment that explains why we disable the cache for this types?

Issue #579, opened 2024-01-24
Telemetry CR in UI


strekm:
**Description**

Add UI extension for Telemetry CRD.

ACs:
- [x] it is possible to manage Telemetry CR using Kyma dashboard
- [x] UI is tested with automated test

**Reasons**

Improve UX of Istio module

**Attachments**
https://kyma-project.io/#/istio/user/operation-guides/02-30-enable-istio-access-logs
https://istio.io/latest/docs/reference/config/telemetry/
https://github.com/kyma-project/istio/tree/main/config/ui-extensions

**PR**
- https://github.com/kyma-project/istio/pull/758

a-thaler:
I assume not all options should be supported in the form view as they are not part of any documented use case. We should select wisely and promote the settings important for our flows like the simple activation of access logs using the predefined provider (not any custom provider)

PR #856, opened 2024-06-03
fix: Use a custom path for telemetries UI to avoid conflicts with kyma telemetry resources

a-thaler:
**Description**
With https://github.com/kyma-project/istio/issues/579 a new UI for istio telemetry resources was introduced. The UI is in conflict with Kyma telemetry resources as they share the same URL path.

With https://github.com/kyma-project/telemetry-manager/pull/1092 we tried to change the path for the kyma resources. The actual navigation node is fixed indeed, however, the module UI of the dashboard does ignore that setting and build up links on the plural of the CRD and is still broken.

As fixing the dashboard will take time, this PR is adjusting the path for the istio resource as they are not used in the module UI, so that the kyma resource path can be reverted back.

Screenshot 2024-06-03 at 14 19 21

- change path from "telemetries" to "istiotelemetries" for telemetry UI

**Related issues**

### DIFF
--- a/config/ui-extensions/telemetries/general
+++ b/config/ui-extensions/telemetries/general
@@ -2,7 +2,7 @@ resource:
   kind: Telemetry
   group: telemetry.istio.io
   version: v1alpha1
-urlPath: telemetries
+urlPath: istiotelemetries
 category: Istio
 name: Telemetries
 scope: namespace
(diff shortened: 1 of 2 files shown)


a-thaler:
Probably blocked by https://github.com/kyma-project/busola/issues/2927

a-thaler:
I talked to @mrCherry97 and we aligned that he will fix https://github.com/kyma-project/busola/issues/2927, after that the UI test should get green and we can go forward

barchw (approved):

barchw (approved):

Issue #867, opened 2024-06-05
Istio 1.21.3 update


strekm:
**Description**

Update the Kyma Istio module to use new Istio version. Update Istio version and used dependencies, adjust tests and documentation if needed. The upgrade needs to have zero downtime for production settings.

ACs:
- [x] Review Istio RNs.
- [x] Verify that the new Istio version doesn't introduce features that transition to a new [phase](https://istio.io/latest/docs/releases/feature-stages/) in Istio, potentially affecting Kyma's Istio behavior. If such changes are identified, discuss them with the team to determine the best course of action.
- [x] Istio bumped on the `main` and latest release branch.
- [x] Prepare Kyma runtime Istio RNs based on open-source Istio RNs.
- [x] Istio installs and upgrades to new version.
- [x] Istio module upgrades with zero downtime - https://github.com/kyma-project/istio/issues/429
- [x] Tests and documentation updated if needed.
- [x] Verify that sidecars are in sync with Control Plane.
- [x] Istio and Envoy Version updated in the [`README.md`](https://github.com/kyma-project/istio) and [`/docs/user/README.md`](https://github.com/kyma-project/istio) files. You can use the scripts `scripts/get_module_istio_version.sh` and `scripts/get_module_envoy_version.sh` to extract the versions.

**DoD:**
- [x] Provide documentation.
- [x] Test on a production-like environment.
- [x] Verify if the solution works for both open-source Kyma and SAP BTP, Kyma runtime.
- [x] Check the outcome of all related pipelines.
- [x] As a PR reviewer, verify code coverage and evaluate if it is acceptable.
- [x] Add release notes.

**Attachments**
[Istio Release Announcements](https://istio.io/latest/news/releases/1.21.x/announcing-1.21.3/)

**PRs**:
- [x] https://github.com/kyma-project/istio/pull/877 (main)
- [x] https://github.com/kyma-project/istio/pull/882 (release-1.6)
- [x] https://github.com/kyma-project/test-infra/pull/10986

PR #877, opened 2024-06-11
Bump Istio to 1.21.3

videlov:
**Description**

- Bump Istio to 1.21.3

**Related issues**

https://github.com/kyma-project/istio/issues/867


docs/release-notes/1.6.4.md
```diff
@@ -0,0 +1,3 @@
+## New Features
```
barchw, line 1: Previously patch release notes weren't added on main branch, only on release

barchw (changes requested):

docs/release-notes/1.6.4.md
videlov: tbh I would like to have everything on main and distributed to patch release branches, but lets keep it consistent like until now.

barchw (approved):

PR #896, opened 2024-06-18
Grpc loadtest

barchw:
**Description**

- ...

**Related issues**


Makefile
```diff
@@ -217,6 +217,12 @@ istio-integration-test: install deploy
 	# Increased TEST_REQUEST_TIMEOUT to 300s to avoid timeouts on newly created k3s clusters
 	cd tests/integration && TEST_REQUEST_TIMEOUT=300s && EXPORT_RESULT=true go test -v -timeout 35m -run TestIstioMain
 
+.PHONY: grpc-performance-test
+grpc-performance-test: install deploy
```
triffer, line 221: Should we use install deploy here? Since this will not work out of the box as the image in the config will not work on Gardener. Also the istio module would not be installed on the cluster, because Istios default resource is not created.

I'd remove this make target and just use the documentation and targets in `tests/performance-grpc`.

tests/performance-grpc/README.md
```diff
@@ -0,0 +1,32 @@
...
+    make deploy-helm
+    ```
+
+2. Run the test with the following command:
+
+    ```bash
+    make grpc-load-test
+    ```
+
+3. To get the tests results run the following command:
```
triffer, line 23: Maybe some info where test results are exported to will be helpful.

triffer (commented):
I was able to run the tests on my Gardener cluster, looks good. 👍 

We don't explicitly control the scaling of ingressgateway or maybe I missed this. I remember that we discussed about this for performance tests.

tests/performance-grpc/steps/job.yaml
```diff
@@ -0,0 +1,28 @@
...
+  template:
+    metadata:
+      labels:
+        sidecar.istio.io/inject: "false"
+    spec:
+      containers:
+      - name: loadtest
+        image: "europe-central2-docker.pkg.dev/sap-se-cx-kyma-goat/networking-dev-tools/grpc-loadtest:latest"
+        command:
+        - /manager
```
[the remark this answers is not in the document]
barchw: Not really, just wanted to provide the arguments
Guess I could have just used `args` instead

tests/performance-grpc/steps/steps.go
```diff
@@ -0,0 +1,123 @@
...
+		return err
+	}
+
+	name := string(uuid.NewUUID())
+	t.templatedValues["HTMLReportPath"] = fmt.Sprintf("%s.html", name)
+
+	useExternalIngressGatewayHost := os.Getenv("USE_EXTERNAL_INGRESS_GATEWAY_HOST") == "true"
+	if useExternalIngressGatewayHost {
+		host := getRequestHost()
+		if host != "" {
```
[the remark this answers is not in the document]
barchw: If the host is empty, the default value from the job.yaml template will be used
I don't see an issue here, with this you can also run the test calling the svc directly

There's a print for this in `getRequestHost()`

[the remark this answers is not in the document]
barchw: OK, you might be right now that I think of it :D
The original reason behind this was to support calling the gRPC service directly, but then I changed the default address to `istio-ingressgateway.istio-system.svc.cluster.local:443`
It might be better than to change the default to `grpcbin:50051`, so the test would also support running it without istio

Makefile
barchw, line 221:
```suggestion
grpc-performance-test:
```

tests/performance-grpc/steps/job.yaml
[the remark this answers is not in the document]
barchw: Ah the naming! Yes, than it's a copy paste :D

triffer (dismissed):

tests/performance-grpc/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+    ```bash
+    make export-results
+    ```
+   The results are stored in the `results` directory in an HTML format.
+
+## Test configuration
```
nataliasitko, line 38:
```suggestion
## Test Configuration
```

tests/performance-grpc/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+[Feature files](./features) allow configuration of command line arguments for [grpc-loadtest](https://github.com/kyma-project/networking-dev-tools/tree/main/grpc-loadtest) tool. 
+Each scenario is deployed as a separate Job. The job template is stored in [here](./steps/job.yaml).
```
nataliasitko, line 41:
```suggestion
Each scenario is deployed as a separate Job. The job template is stored in [`/steps/job.yaml`](./steps/job.yaml).
```

tests/performance-grpc/README.md
```diff
@@ -0,0 +1,41 @@
...
+## Running the test on AWS
+
+Out of the box, the test does not support running on AWS, as the `deploy-helm` Makefile target gets the external IP of the Istio Ingress Gateway.
+The external IP is used in the `grpc-loadtest` command to send requests to the service.
+The external IP is not available on AWS, so the test will fail.
+To run the test on AWS, you need to deploy the helm chart with the external address that AWS provides on the LoadBalancer service.
+
+## Running the test
+
+1. Install the helm chart with the following command:
```
nataliasitko, line 19:
```suggestion
1. Install the helm chart using the following command:
```

tests/performance-grpc/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+    ```bash
+    make deploy-helm
+    ```
+
+2. Run the test with the following command:
```
nataliasitko, line 25:
```suggestion
2. Run the test using the following command:
```

tests/performance-grpc/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+    ```bash
+    make grpc-load-test
+    ```
+
+3. To get the tests results run the following command:
```
nataliasitko, line 31:
```suggestion
3. To get the test's results, run the following command:
```

tests/performance-grpc/README.md
```diff
@@ -0,0 +1,41 @@
...
+To provide a valuable test result, the Istio module configuration should ensure that istio-ingress gateway has a constant number of replicas, and is not scaled up or down during the test.
+
+## Running the test on AWS
+
+Out of the box, the test does not support running on AWS, as the `deploy-helm` Makefile target gets the external IP of the Istio Ingress Gateway.
+The external IP is used in the `grpc-loadtest` command to send requests to the service.
+The external IP is not available on AWS, so the test will fail.
+To run the test on AWS, you need to deploy the helm chart with the external address that AWS provides on the LoadBalancer service.
+
+## Running the test
```
nataliasitko, line 17:
```suggestion
## Running the Test
```

nataliasitko (approved):

PR #906, opened 2024-06-27
Remove dependency to Prow in Pull Request workflows

triffer:
**Description**

- Replace waiting for Prow job with building a Docker image and using it in the k3d integration tests in the Pull Request workflow
- Replace waiting for Prow job with building a Docker image and pushing it to GCP Artifact Registry in Gardener-based PR integration tests

**Related issues**

### DIFF
--- a/.github/actions/build-manager-image/action.yaml
+++ b/.github/actions/build-manager-image/action.yaml
@@ -1,0 +1,64 @@
+name: 'Build Istio Manager Image'
+description: 'Builds Istio manager image'
+inputs:
+  operator-image-name:
+    description: 'Operator image name used for creation of Istio manager image'
+    required: true
+  platforms:
+    description: 'Platforms to build the image for (comma-separated), e.g. linux/amd64,linux/arm64,linux/arm/v7'
+    required: false
+    default: 'linux/amd64'
+  push-image:
+    description: 'Push the image to GCP'
+    required: false
+    default: 'false'
+  push-sa-key:
+    description: 'Service account key to push the image to GCP'
+    required: false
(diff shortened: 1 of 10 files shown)


.github/actions/build-manager-image/action.yaml
```diff
@@ -0,0 +1,60 @@
+name: 'Build Istio Manager Image'
+description: 'Builds Istio manager image'
+inputs:
+  operator-image-name:
+    description: 'Operator image name used for creation of Istio manager image'
+    required: true
+  platforms:
```
kolodziejczak, line 7: After dropping ARM build idea, can this input be removed?

.github/actions/build-manager-image/action.yaml
```diff
@@ -0,0 +1,60 @@
...
+    required: false
+    default: 'false'
+  push-sa-key:
+    description: 'Service account key to push the image to GCP'
+    required: false
+    default: ''
+runs:
+  using: "composite"
+  steps:
+    - name: Set up QEMU
```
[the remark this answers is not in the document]
triffer: We don't need it and I kept it, just if we want to use it in the future, since it only takes ~2 seconds. But I can remove it, if you think it is better to remove it. But then I'd also remove the platform arg and always build it for linux.

.github/actions/build-manager-image/action.yaml
triffer: Yes, probably if we agree on, that we don't want to support it, I'd remove it together with QEMU.

nataliasitko (approved):

Issue #921, opened 2024-07-04
Rate limit technical design


strekm:
**Description**

Based on previous POCs and [decision](https://github.com/kyma-project/istio/issues/773#issuecomment-2208182818) to support local rate limiting as part of Istio module provide technical design and API design. This should be focused on use case of local rate limiting but having in mind that later on global rate limiting might be introduced. Technical design should specify additional controller, still being part of Istio module, that will be listening to rate limit CRD.

Main scenario to cover with local rate limit is to protect a workload from overloading. It should be possible to configure number of requests per workload. 

ACs:
- [x] technical design of rate limit functionality
- [x] API design for local rate limiting 
- [x] above should be extensible to global rate limiting
- [x] ADR created

**Reasons**

**DoD:**
~~- [ ] Provide unit and integration tests.~~
- [x] Provide documentation.
~~- [ ] Verify if the solution works for both open-source Kyma and SAP BTP, Kyma runtime.~~
~~- [ ] If you changed the resource limits, explain why it was needed.~~
~~- [ ] If the default configuration of Istio Operator has been changed, you performed a manual upgrade test to verify that the change can be rolled out correctly.~~
~~- [ ] Verify that your contributions don't decrease code coverage. If they do, explain why this is the case.~~
~~- [ ] Add release notes.~~

**Attachments**
POCs: https://github.com/kyma-project/istio/issues/774, https://github.com/kyma-project/istio/issues/826
part of: #773

PR:  https://github.com/kyma-project/istio/pull/930

PR #930, opened 2024-07-11
Rate limit technical design

kolodziejczak:
**Description**

- Technical design of rate limit functionality
- ADR
- API design

**Related issues**

#921


docs/contributor/draft/rate-limit-api.yaml
```diff
@@ -0,0 +1,54 @@
+# Since combining global and global rate limit would make API more complex
```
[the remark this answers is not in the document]
kolodziejczak: thanks for catching that!

docs/contributor/draft/rate-limit-api.yaml
```diff
@@ -0,0 +1,54 @@
...
+kind: LocalRateLimit
+metadata:
+  name: httpbin-local-rate-limit
+  namespace: default
+spec:
+  workloadSelector:
+    labels:
+      app: httpbin
+  rateLimitBy:
+    requests:
```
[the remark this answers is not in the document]
kolodziejczak: let's go simple at the beginning, and leave the room to extend it with additional option to specify buckets. I'll adjust it this way

docs/contributor/draft/rate-limit-api.yaml
```diff
@@ -0,0 +1,54 @@
+# Since combining global and global rate limit would make API more complex
+# I would avoid doing so, and in case of extending Istio Module capabilities
+# to global rate limit, I'd introduce new CRD. This way we would make use of our API easier.
+# User would create LocalRateLimit and GlobalRateLimit custom objects separately
+# One LocalRateLimit per workload for which rate limit is applied.
+# This way it's easier for user to keep service related resources in one place.
+# Also avoiding long yamls in case of big clusters with big amounts of workloads
+
+# This is how we could introduce it for now covering simple rate limiting without any specific selectors
+apiVersion: operator.kyma-project.io/v1alpha1
```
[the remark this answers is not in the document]
kolodziejczak: Sure, this is just an example to align if it's going into the right direction. I'll create more specified view of CRD

docs/contributor/draft/rate-limit-api.md
```diff
@@ -0,0 +1,122 @@
...
+kind: RateLimit
+metadata:
+  name: default
+  namespace: kyma-system
+spec:
+  workloadSelector:
+    labels:
+      app: httpbin
+  localRateLimit: # separating it like this allows to extend fields with new use cases. Also every field can be extended with it's specifics
+    localRateLimit:
```
barchw, line 57: The field name is doubled here (`localRateLimit.localRateLimit`)

docs/contributor/draft/rate-limit-api.md
```diff
@@ -0,0 +1,122 @@
+# Rate Limit Proposal
+
+## Spec 
+
+| field                                            | type        | description                                                     | required                      |
+|--------------------------------------------------|-------------|-----------------------------------------------------------------|-------------------------------|
+| wokloadSelector                                  | object      | in case of empty, it's applied for the whole namespace          | no                            |
+| workloadSelector.labels                          | object      | allows to select given workload by a selector                   | no                            |
+| localRateLimit                                   | object      | allows to describe local rate limit properties                  | yes                           |
+| localRateLimit.by                                | object      | container for criteria you can limit by (for the future)        | yes                           |
```
barchw, line 10: Any reason to have that `by` field? 
Wouldn't `localRateLimit.requests` or `localRateLimit.headerPresence` be equivalent?

docs/contributor/draft/rate-limit-api.md
```diff
@@ -0,0 +1,122 @@
...
+| localRateLimit.by.requests                       | object      | allows to set rate limit based on plain number of requests      | yes                           |
```
barchw, line 11: Was support for array considered? After all envoy does support [actions as an array].(https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/rate_limit_filter#composing-actions)
Example (but only with one): https://github.com/kyma-project/istio/blob/88a1b253bf08da601a239ca66fe18afd1169b910/docs/contributor/draft/rate-limiting-poc/global/scenario-1-limit-by-dynamic-header-values.yaml#L39

barchw (changes requested):

docs/contributor/draft/rate-limit-api.md
kolodziejczak: would be, I put it here since it felt nice to have it readable like a sentence

docs/contributor/draft/rate-limit-api.md
kolodziejczak: oh thanks

docs/contributor/draft/rate-limit-api.md
kolodziejczak: do you think dropping this and avoiding unnecessary nesting would be worth more than this?

docs/contributor/draft/rate-limit-api.md
kolodziejczak: Since we abstract EnvoyFIlter away and control directly which of the rate limit criterias are exposed, I feel like it's easier to use direct fields instead of arrays.

docs/contributor/draft/rate-limit-api.md
barchw: Yes, this would be more similar to the strategy we took for API Gateway handler (ext auth, jwt, noauth)

docs/contributor/draft/rate-limit-api.md
barchw: It's easier now; If, for some reason we would decide to switch to using array in the future, this would come with a breaking change. That's why I'm asking about whether it was considered.

kolodziejczak: I wouldn't worry about changing it to array in the future. For now this approach is easier, fits well and in case of need to go into array, we can introduce new version of the CRD.

docs/contributor/draft/rate-limit-api.md
```diff
@@ -0,0 +1,121 @@
+# Rate Limit Proposal
+
+## Spec 
+
+| field                                            | type        | description                                                     | required                      |
+|--------------------------------------------------|-------------|-----------------------------------------------------------------|-------------------------------|
+| wokloadSelector                                  | object      | in case of empty, it's applied for the whole namespace          | no                            |
```
triffer, line 7: I think there was also proposed that the RateLimit CR will behave the same like EnvoyFilter, e.g. to apply to a namespace, or all namespaces, depending where it is applied. Information can be found in EnvoyFilter workloadSelector documentation: https://istio.io/latest/docs/reference/config/networking/envoy-filter/#EnvoyFilter

> If omitted, the set of patches in this configuration will be applied to all workload instances in the same namespace. If the EnvoyFilter is present in the config root namespace, it will be applied to all applicable workloads in any namespace.

docs/contributor/draft/rate-limit-api.md
```diff
@@ -0,0 +1,121 @@
...
+        max_tokens: 1000
+        tokens_per_fill: 1000
+        fill_interval: 1s
+      etc: "etc"
+  globalRateLimit:
+    requests:
+      time:
+        unit: "second" 
+        requestsLimit: 30
+      buckets:
```
triffer, line 103: Global rate limiting doesn't support buckets.

docs/contributor/draft/rate-limit-api.md
```diff
@@ -0,0 +1,121 @@
...
+| localRateLimit.by.requests.time.unit             | string      | describes unit of time for rate limit                           | yes                           |
+| localRateLimit.by.requests.time.requestsLimit    | int         |describes requests limit for a declared time unit                | yes                           |
+| localRateLimit.by.requests.buckets.maxTokens     | int         |describes max amount of tokens that can stack up                 | yes                           |
+| localRateLimit.by.requests.buckets.tokensPerFill | int         | describes how many tokens to fill every defined period of time  | yes                           |
+| localRateLimit.by.requests.buckets.fillInterval  | time        |  describes how often fill tokens                                | yes                           |
+
+## Usage example
+
+```yaml
+apiVersion: operator.kyma-project.io/v1alpha1
```
triffer, line 23: It would be good to also have an example to configure rate limit for ingress gateway, since those are basically the two scenarios where rate limit can be applied: on workload and on ingress gateway.

docs/contributor/draft/rate-limit-api.md
triffer, line 1: Did you check if we need to expose the [PatchContext](https://istio.io/latest/docs/reference/config/networking/envoy-filter/#EnvoyFilter-PatchContext) in the rate limitCR? What is used in the EnvoyFilter if we don't set it? Is it still working as expected for Gateway and Sidecar?

docs/contributor/draft/rate-limit-api.md
```diff
@@ -0,0 +1,121 @@
...
+| wokloadSelector                                  | object      | in case of empty, it's applied for the whole namespace          | no                            |
+| workloadSelector.labels                          | object      | allows to select given workload by a selector                   | no                            |
+| localRateLimit                                   | object      | allows to describe local rate limit properties                  | yes                           |
+| localRateLimit.by                                | object      | container for criteria you can limit by (for the future)        | yes                           |
+| localRateLimit.by.requests                       | object      | allows to set rate limit based on plain number of requests      | yes                           |
+| localRateLimit.by.requests.time                  | object      | allows to describe local rate limiting using time periods       | yes (if buckets not defined)  |
+| localRateLimit.by.requests.buckets               | object      | allows to describe local rate limiting using buckets            | yes (if time not defined)     |
```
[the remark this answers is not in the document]
kolodziejczak: right, better avoid plural form here

docs/contributor/draft/rate-limit-api.md
```diff
@@ -0,0 +1,121 @@
...
+|--------------------------------------------------|-------------|-----------------------------------------------------------------|-------------------------------|
+| wokloadSelector                                  | object      | in case of empty, it's applied for the whole namespace          | no                            |
+| workloadSelector.labels                          | object      | allows to select given workload by a selector                   | no                            |
+| localRateLimit                                   | object      | allows to describe local rate limit properties                  | yes                           |
+| localRateLimit.by                                | object      | container for criteria you can limit by (for the future)        | yes                           |
+| localRateLimit.by.requests                       | object      | allows to set rate limit based on plain number of requests      | yes                           |
+| localRateLimit.by.requests.time                  | object      | allows to describe local rate limiting using time periods       | yes (if buckets not defined)  |
```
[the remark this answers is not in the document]
kolodziejczak: Let's start with exposing only time then, it won't be eventually a lot of work in the future to expose bucket approach in case of any feature request.

docs/contributor/draft/rate-limit-api.md
kolodziejczak: right, config root ns missed here

docs/contributor/draft/rate-limit-api.md
kolodziejczak: To apply rate limit on ingress proxies we need to apply PatchContext `GATEWAY`, to rate limit on sidecars inbound traffic listeners we need `SIDECAR_INBOUND`. I assume we don't want to rate limit outbound traffic from the sidecars. In case of not specifying it, `ANY` should be applied, which also will limit the outbound traffic from sidecars.

kolodziejczak: I think we can abstract away PatchContext from users to a bit simpler form like example: `applyTo: "ingress" | "sidecars"`

docs/contributor/adr/0007-rate-limit.md
```diff
@@ -0,0 +1,28 @@
...
+## Context
+
+It has been decided to indroduce rate limit functionality into Kyma platform to allow users to set rate limit functionality on the service mesh layer, therefore allowing to consume intended service mesh functionality which is abstracting away networking concerns outside applications inside the mesh.
+Since Istio is an underlying service mesh responsible for the workloads networking across Kyma clusters, therefore it is the Istio which allows to configure such functionality.
+Taking that into account Goat Team is responsible to introduce this feature.
+At this moment it is not possible to introduce global rate limit provided by Istio service mesh because of technical or legal limitations related to the deployment of key-value stores in the cluster.
+
+## Decision
+
+Taking the context part into account, it was decided to implement local rate limit as the only rate limiting posibility for now.
```
mluk-sap, line 16: Does it make sense to explain here what does the local mean in this context (implicitly opposed to 'global')?

docs/contributor/draft/rate-limit-api.md
kolodziejczak: As we agreed, let's go with `context: "gateway" | "sidecar"`

docs/contributor/adr/0007-rate-limit.md
kolodziejczak: Sure, it's a good idea since it can be confusing sometimes. I'll try to add some short explanation.

docs/contributor/draft/rate-limit-api.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Spec 
+
+| field                                            | type        | description                                                     | required                      |
+|--------------------------------------------------|-------------|-----------------------------------------------------------------|-------------------------------|
+| wokloadSelector                                  | object      | in case of empty, it's applied for the whole namespace          | no                            |
+| workloadSelector.labels                          | object      | allows to select given workload by a selector                   | no                            |
+| localRateLimit                                   | object      | allows to describe local rate limit properties                  | yes                           |
+| localRateLimit.by                                | object      | container for criteria you can limit by (for the future)        | yes                           |
+| localRateLimit.by.requests                       | object      | allows to set rate limit based on plain number of requests      | yes                           |
```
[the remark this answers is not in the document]
kolodziejczak: Reiterated over that, api proposal adjusted.

docs/contributor/draft/rate-limit-api.yaml
```diff
@@ -0,0 +1,54 @@
...
+kind: LocalRateLimit
+metadata:
+  name: httpbin-local-rate-limit
+  namespace: default
+spec:
+  workloadSelector:
```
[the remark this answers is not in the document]
kolodziejczak: In case of workloadSelector being skipped it applies to all workloads within the namespace, and in addition when it's in the istio-system (root config ns) it applies to all workloads across all namespaces. Similar to the EnvoyFilter way of working.

docs/contributor/adr/0007-rate-limit.md
```diff
@@ -0,0 +1,30 @@
...
+Since Istio is an underlying service mesh responsible for the workloads networking across Kyma clusters, therefore it is the Istio which allows to configure such functionality.
+At this moment it is not possible to introduce global rate limit provided by Istio service mesh because of the technical or legal limitations, related to the deployment of key-value stores in the cluster.
+Major difference between local and global rate limit is that local rate limiting is applied directly at the sidecar proxy level, controlling the rate of requests on a per-instance basis. Global rate limiting is applied accross multiple instances of a service or across, multiple services, providing a coorinated rate limit within the mesh.
+In case of introduction the global rate limit functionality in the future, new ADR will be created.
+
+## Decision
+
+It was decided to implement local rate limit as the only rate limiting option at this time.
```
triffer, line 17: I thought we wanted to mention that the global rate limiting is out of scope of this ADR, because of the limitations? 🤔  In this case I'd not mention it as a decision.

docs/contributor/adr/0007-rate-limit.md
```diff
@@ -0,0 +1,30 @@
...
+New CRD in the Istio module will be introduced to allow its configuration.
```
triffer, line 18:
```suggestion
A new CRD will be introduced in the Istio module to allow rate limiting configuration.
```

docs/contributor/adr/0007-rate-limit.md
```diff
@@ -0,0 +1,30 @@
...
+Currently the possibilites for the rate limit functionality are limited to restrict number of requests, without any other criteria like header presence or header value. 
+CRD is structured in a way that allows further extensions, allowing to configure different rate limit criteria, or to introduce global rate limit.
+
+## Consequences
+
+New controller for the new rate limit CRD needs to be implemented as a part of the Istio module.
+The new rate limit CR can be used by users to set rate limits in the service mesh in their cluster without having to worry about possible changes in the Istio EnvoyFilter resources.
+Nevertheless EnvoyFilter is complex and since it's API is not stable yet, current RateLimit CR will be provided as v1alpha1 version, and might be changed in the future Istio releases.
+RateLimit CRD should be included into blocking deletion strategy for IstioCR, since Istio should not be uninstalled while RateLimit CRs exist in the cluster.
```
triffer, line 28:
```suggestion
RateLimit CRD should be included into blocking deletion strategy for Istio CR, since Istio should not be uninstalled while RateLimit CRs exist in the cluster.
```

docs/contributor/adr/0007-rate-limit.md
```diff
@@ -0,0 +1,30 @@
...
+New functionality requires integration tests to be implemented.
+In case of an introduction of any additional criteria to the local rate limit in the RateLimit CR, the capacity of rate limit on this criteria should be lower than or at most equal to the root setting. Otherwise it will be reduced to the root capacity value.
```
triffer, line 30: I think it might be hard to understand that. Maybe you can be a bit more descriptive and provide more context?

docs/contributor/draft/rate-limit-design.md
```diff
@@ -0,0 +1,80 @@
+# Rate Limit Proposal
+
+## Spec 
+
+| field                                            | type        | description                                                     | required                      |
+|--------------------------------------------------|-------------|-----------------------------------------------------------------|-------------------------------|
+| wokloadSelector                                  | object      | when skipped, it's applied for the whole namespace workloads, if empty and in istio-system ns, applies to all workloads in all namespaces | no                            |
```
triffer, line 7: I'd reuse documentation from Istio.
```suggestion
| workloadSelector                                  | object      | If omitted, the rate limit will be applied to all workload instances in the same namespace as the RateLimit CR. If the RateLimit CR is present in the istio-system namespace, it will be applied to all workloads in any namespace. | no                            |
```

docs/contributor/draft/rate-limit-design.md
```diff
@@ -0,0 +1,80 @@
...
+| workloadSelector.labels                          | object      | allows to select given workload by a label selector                   | no                            |
```
triffer, line 8: Why are we not reusing the expressive Istio documentation?
I think the type is not object but map<string, string>.
```suggestion
| workloadSelector.labels                          | map<string, string>      | One or more labels that indicate a specific set of pods/VMs on which the configuration should be applied. The scope of label search is restricted to the configuration namespace in which the the resource is present.                  | no                            |
```

docs/contributor/draft/rate-limit-design.md
```diff
@@ -0,0 +1,80 @@
...
+| local                                   | object      | allows to describe local rate limit properties                  | yes                           |
```
triffer, line 9: Should we have it as required? If we introduce global rate limiting it might not be required any longer, so we will have to change this validation once we introduce global rate limiting. If this is fine for us, we need to mention this in the decision and consequences of the ADR.

docs/contributor/adr/0007-rate-limit.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Decision
+
+It was decided to implement local rate limit as the only rate limiting option at this time.
+New CRD in the Istio module will be introduced to allow its configuration.
+In the future, global rate limit might be introduced. The decision has not been made yet, therefore it has been decided to name the new CRD `RateLimit` to not to indicate, that global rate limit will be provided. By naming current CRD for example `LocalRateLimit` logical would be to expect also `GlobalRateLimit` to be existent at some point.
+Currently the possibilites for the rate limit functionality are limited to restrict number of requests, without any other criteria like header presence or header value. 
+CRD is structured in a way that allows further extensions, allowing to configure different rate limit criteria, or to introduce global rate limit.
+
+## Consequences
```
triffer, line 23: Maybe we should add a consequence that rate limited requests will result in a status code 429?

docs/contributor/adr/0007-rate-limit.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Context
+
+It has been decided to indroduce rate limit functionality into Kyma platform to allow users to set rate limit functionality on the service mesh layer, therefore allowing to consume intended service mesh functionality which is abstracting away networking concerns outside applications inside the mesh.
+Since Istio is an underlying service mesh responsible for the workloads networking across Kyma clusters, therefore it is the Istio which allows to configure such functionality.
+At this moment it is not possible to introduce global rate limit provided by Istio service mesh because of the technical or legal limitations, related to the deployment of key-value stores in the cluster.
+Major difference between local and global rate limit is that local rate limiting is applied directly at the sidecar proxy level, controlling the rate of requests on a per-instance basis. Global rate limiting is applied accross multiple instances of a service or across, multiple services, providing a coorinated rate limit within the mesh.
+In case of introduction the global rate limit functionality in the future, new ADR will be created.
+
+## Decision
```
triffer, line 15: Did we think about enforcement: https://github.com/kyma-project/istio/blob/rate-limit-poc/docs/contributor/draft/rate-limiting-poc/01-rate-limiting.md#rate-limit-enforcement?

triffer, line 15: Do we want to return the rate limit headers per default?
https://github.com/kyma-project/istio/blob/rate-limit-poc/docs/contributor/draft/rate-limiting-poc/01-rate-limiting.md#rate-limit-response-headers

docs/contributor/adr/0007-rate-limit.md
triffer, line 23: Do we want to have a consequence about documentation of rate limit metrics for user enablement?
https://github.com/kyma-project/istio/blob/rate-limit-poc/docs/contributor/draft/rate-limiting-poc/01-rate-limiting.md#rate-limit-metrics

docs/contributor/draft/rate-limit-design.md
```diff
@@ -0,0 +1,80 @@
...
+| local.bucket                            | object      | allows to set rate limit capacity through bucket concept        | yes                           |
```
triffer, line 10:
```suggestion
| local.bucket                            | object      | The token bucket configuration to use for rate limiting requests. Each request consumes a single token. If the token is available, the request will be allowed. If no tokens are available, the request will be rejected with status code 429.        | yes                           |
```

docs/contributor/draft/rate-limit-design.md
```diff
@@ -0,0 +1,80 @@
...
+| local.bucket.max_tokens                 | int         | max tokens that can be gathered for the rate limit capacity     | yes                           |
```
triffer, line 11:
```suggestion
| local.bucket.max_tokens                 | int         | The maximum tokens that the bucket can hold. This is also the number of tokens that the bucket initially contains.  | yes                           |
```

docs/contributor/draft/rate-limit-design.md
```diff
@@ -0,0 +1,80 @@
...
+| local.bucket.tokens_per_fill            | int         | number of tokens added per fill cycle                           | yes                           |
```
triffer, line 12:
```suggestion
| local.bucket.tokens_per_fill            | int         | The number of tokens added to the bucket during each fill interval.   | yes                           |
```

docs/contributor/draft/rate-limit-design.md
```diff
@@ -0,0 +1,80 @@
...
+| local.bucket.fill_interval              | time        | interval on which tokens are filled                             | yes                           |
```
triffer, line 13: We should also have a consequence that we have validation for the minimum fill_interval value. Documentation about this can be found here: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/local_ratelimit/v3/local_rate_limit.proto#local-rate-limit-proto
```suggestion
| local.bucket.fill_interval              | time        | The fill interval that tokens are added to the bucket. During each fill interval `tokens_per_fill` are added to the bucket. The bucket will never contain more than `max_tokens` tokens. The `fill_interval` must be >= 50ms to avoid too aggressive refills.                            | yes                           |
```

triffer, line 13: I'm not sure if it should be duration or time? I remember that we used `duration` in APIRule v1beta1 for CORS maxAge.
```suggestion
| local.bucket.fill_interval              | duration        | interval on which tokens are filled                             | yes                           |
```

docs/contributor/draft/rate-limit-design.md
```diff
@@ -0,0 +1,80 @@
+# Rate Limit Proposal
+
+## Spec
```
triffer, line 3: Where is the `context` field in the spec?

docs/contributor/draft/rate-limit-design.md
```diff
@@ -0,0 +1,80 @@
...
+## Usage example
+
+```yaml
+apiVersion: operator.kyma-project.io/v1alpha1
+kind: RateLimit
+metadata:
+  name: httpbin-local-rate-limit
+  namespace: default
+spec:
+  context: "gateway" | "sidecar" # translates to: PatchContext: GATEWAY | SIDECAR_INBOUND, should be on this level to not to mix sidecar level rate limit with gateway level rate limit. In such case, 2 separate resources should be created.
```
triffer, line 24: In case gateway context is set, do we do some magic to apply it automatically to `istio-system` and `ingressgateway` or do we want to be expressive? I think having less "magic" for a starter might be better. We might want to mention this in the decision.

triffer, line 24: If this is an example we should have here context: "sidecar" and not both values.

docs/contributor/draft/rate-limit-design.md
```diff
@@ -0,0 +1,80 @@
...
+      app: httpbin
+  local:
+    bucket:
+      max_tokens: 1000
+      tokens_per_fill: 1000
+      fill_interval: 1s
+
+---
+
+# eventually we could extend it later into such direction
```
triffer, line 36: If we have already an ideation for upcoming fields, we should also briefly mention what fields we are considering, e.g.
- **spec.local.time** to simplify configuration of buckets.
- **spec.local.additionalCriteria** ...
- ...

docs/contributor/draft/rate-limit-design.md
```diff
@@ -0,0 +1,80 @@
...
+          fill_interval: 1s
+        headerName: "some-header"
+        headerValue: "some-value"
+        vhost: "some.v.host"
+        route: "/some/path"
+  global:
+    globalSpecificSettings:
+```
+
+Additional details:
```
triffer, line 73: Do we need this, because it should rather be mentioned in the ADR.

docs/contributor/draft/rate-limit-design.md
```diff
@@ -0,0 +1,80 @@
...
+- Blocking deletion strategy for IstioCR should be extended with RateLimit CRD
+- In case of need to apply RateLimit CR to all namespaces, it should be placed in the istio-system, mimicking EnvoyFilter strategy for global application. Therefore underlying EnvoyFilter should be placed in the istio-system namespace as well. In such scenario spec.workloadSelector should not be present.
```
triffer, line 75: We should have this in the ADR as well.

docs/contributor/draft/rate-limit-design.md
kolodziejczak: just added, thanks!

docs/contributor/adr/0007-rate-limit.md
```diff
@@ -0,0 +1,30 @@
...
+Since Istio is an underlying service mesh responsible for the workloads networking across Kyma clusters, therefore it is the Istio which allows to configure such functionality.
+In Envoy there is support for local and global rate limiting. The major difference between local and global rate limit is that local rate limiting is applied directly at the sidecar proxy level, controlling the rate of requests on a per-instance basis. 
+For consistent rate limiting across your service mesh, the global rate limiting can be utilised. The global rate limiting requires a rate limit service that adheres to the [gRPC RateLimit v3 protocol](https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/ratelimit/v3/rls.proto). This centralised approach enforces shared stateful rate limits, applicable to all instances and services within the mesh.
+Currently, there are limitations that prevent using Redis on managed Kyma clusters. Because global rate limiting relies on Redis for persistence, the global rate limiting is not in scope of this ADR.
+
+## Decision
+
+It was decided to implement local rate limit as the only rate limiting option at this time.
```
kolodziejczak, line 17:
```suggestion
```

docs/contributor/draft/rate-limit-design.md
kolodziejczak: We decided to not do any magic, but to use workload selector value with context to do the validation around it

docs/contributor/draft/rate-limit-design.md
kolodziejczak: let's remove it, as it's outside the scope

docs/contributor/draft/rate-limit-design.md
kolodziejczak: I'll move this to the ADR

docs/contributor/draft/rate-limit-design.md
kolodziejczak: I'll move this to the ADR

docs/contributor/adr/0007-rate-limit.md
kolodziejczak: The only risk I can see in having those headers is that it gives a possibility to fine tune DoS attacks to the rate limit values to make detection harder. Second thing is potential insight into how many requests are getting to the system. I'd turn them off by default as the most secure option, and give users possibility to turn it on with one simple field.

docs/contributor/draft/rate-limit-design.md
kolodziejczak: As we discussed, we are ok with that, and let's add it to the ADR.

docs/contributor/adr/0007-rate-limit.md
triffer: Envoy offers enforcement for 0 - 100% percentage of the traffic. I think the main usecase for enforcement disabled is, that you can check your rate limit metrics before blocking real traffic. Therefore I think it's easier for the user for to have a bool `enforced` to enable or disable enforcement. I think it would be good if this is a required configuration without any default, so the user must make a decision about this.

The following configuration is an example for a rate limit that is enabled, but not enforced:
```yaml
envoy.filters.http.local_ratelimit:
  "@type": type.googleapis.com/udpa.type.v1.TypedStruct
  type_url: type.googleapis.com/envoy.extensions.filters.http.local_ratelimit.v3.LocalRateLimit
  value:
    stat_prefix: api_usage_local_rate_limiter
    enable_x_ratelimit_headers: DRAFT_VERSION_03
    filter_enabled:
      runtime_key: local_rate_limit_enabled
      default_value:
        numerator: 100
        denominator: HUNDRED
    filter_enforced:
      runtime_key: local_rate_limit_enforced
      default_value:
        numerator: 0
        denominator: HUNDRED
    token_bucket:
      max_tokens: 4
      tokens_per_fill: 2
      fill_interval: 60s
```

docs/contributor/draft/rate-limit-design.md
```diff
@@ -0,0 +1,81 @@
...
+| workloadSelector.labels                          | map<string, string>      | One or more labels that indicate a specific set of pods/VMs on which the configuration should be applied. The scope of label search is restricted to the configuration namespace in which the the resource is present.                  | no                            |
+| local                                   | object      | allows to describe local rate limit properties                  | yes                           |
+| local.bucket                            | object      | The token bucket configuration to use for rate limiting requests. Each request consumes a single token. If the token is available, the request will be allowed. If no tokens are available, the request will be rejected with status code 429.        | yes                           |
+| local.bucket.max_tokens                 | int         | The maximum tokens that the bucket can hold. This is also the number of tokens that the bucket initially contains.  | yes                           |
```
triffer, line 12: I just realised that we have camel case and snake case mixed in the CR. We should only use camel case and rename `max_tokens`, `tokens_per_ fill`, ...

docs/contributor/adr/0007-rate-limit.md
triffer: The field would be a boolean `enableHeaders` on the level of `spec`?

docs/contributor/adr/0007-rate-limit.md
triffer: Since the field `stat_prefix` for the local rate limit config in Envoy is required, we need to set a value. This value will be the prefix for the metrics. I think we don't need to make it configurable, yet. Also it might not be required to use a different prefix for each RateLimit configuration. 

My proposal is to use something simple for now for the local rate limiting like `stat_prefix: rate_limit`.
This will result in the following metrics:
```
rate_limit.http_local_rate_limit.enabled: 6
rate_limit.http_local_rate_limit.enforced: 0
rate_limit.http_local_rate_limit.ok: 4
rate_limit.http_local_rate_limit.rate_limited: 2
```

We should add this as a decision.

docs/contributor/adr/0007-rate-limit.md
kolodziejczak: I think we can do enableHeaders. I'll add it to the decision

docs/contributor/adr/0007-rate-limit.md
triffer: I just saw that this behaviour can be controlled with `always_consume_default_token_bucket`([docs](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/local_ratelimit/v3/local_rate_limit.proto#local-rate-limit-proto)). So we should ignore it for now and completely remove it from the consequences. I think this is something to consider when we add more possibilities to local rate limiting.

```suggestion
```

docs/contributor/adr/0007-rate-limit.md
kolodziejczak: Thanks for the work done here, I'll add it to the decision

kolodziejczak: Also I'll do a modification to the API design

docs/contributor/adr/0007-rate-limit.md
kolodziejczak: Sure, I'll try to describe it a bit clearer

docs/contributor/draft/rate-limit-design.md
kolodziejczak: Since IstioCR has all camelCase, I'd go with camelCase here. I think also the convention in k8s CRDs is to go with camelCase, if I'm not wrong.

docs/contributor/adr/0007-rate-limit.md
kolodziejczak: Removed

docs/contributor/adr/0007-rate-limit.md
kolodziejczak: let's remove, the simpler, the better for now

docs/contributor/draft/rate-limit-design.md
kolodziejczak: done

docs/contributor/adr/0007-rate-limit.md
kolodziejczak: thanks for handling that, I'm resolving it

triffer (approved):

docs/contributor/adr/0008-rate-limit.md
```diff
@@ -0,0 +1,103 @@
...
+
+### RateLimit CR Spec
+
+| field                      | type                | description                                                                                                                                                                                                                                               | required |
+|----------------------------|---------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------|
+| context                    | string              | Allows to set context for the underlying EnvoyFilter. Can be "gateway", which means rate limit will be applied to the Istio Ingress Gateways, or "sidecar" which means it will be applied to sidecar proxies of the workloads.                            | yes      |
```
nataliasitko, line 59:
```suggestion
| context                    | string              | Allows to set the context for the underlying EnvoyFilter. The value can be `gateway`, which means the rate limit is applied to the Istio Ingress Gateways, or `sidecar`, which means it is applied to sidecar proxies of the workloads.                            | yes      |
```

docs/contributor/adr/0008-rate-limit.md
```diff
@@ -0,0 +1,103 @@
...
+| enableHeaders              | boolean             | Enable x-rate-limit response headers, the default value is false                                                                                                                                                                                          | yes      |
```
nataliasitko, line 60:
```suggestion
| enableHeaders              | boolean             | Enables **x-rate-limit** response headers. The default value is `false`.                                                                                                                                                                                         | yes      |
```

nataliasitko (approved):

PR #1012, opened 2024-09-10
add workflow for release build with image-builder

kolodziejczak:
**Description**

- add workflow for release build with image-builder


.github/workflows/create-release.yaml
```diff
@@ -85,29 +87,26 @@ jobs:
...
-        name: Verify build of experimental image
-        id: wait-for-experimental-build
+          name: istio-manager
+          dockerfile: Dockerfile
+          context: .
+          build-args: |
+            VERSION=${{ github.event.inputs.name }}
+          tags: "${{ github.event.inputs.name }}"
+
+      - name: Image build - experimental
```
triffer, line 100: Should we use the same name we use on main?
```suggestion
      - name: Build manager image - experimental
```

.github/workflows/create-release.yaml
```diff
@@ -85,29 +87,26 @@ jobs:
         with:
           fetch-depth: 0
 
-      - uses: autotelic/action-wait-for-status-check@6556cf50c8fb6608412945382eae73581f56cbb4
-        name: Build image
-        id: wait-for-build
+      - name: Image build
```
triffer, line 90: Should we use the same name we use on main?
```suggestion
      - name: Build manager image
```

kolodziejczak: looks better, let's do that

.github/workflows/create-release.yaml
kolodziejczak: looks better, let's do that

triffer (approved):

PR #1071, opened 2024-10-11
Remove support for disable external name alias annotation

kolodziejczak:
**Description**

- Remove support for disable external name alias annotation


videlov:
Please update RNs and docs, like in previous PR state:
https://github.com/kyma-project/istio/compare/57ab9156dd473d4ca4be9445c565f5a95a54e511..62a90c89cfc5b2cb32567ef7a8ef9a6c1cc30185

videlov (changes requested):
Please update RNs and docs, like in previous PR state:
https://github.com/kyma-project/istio/compare/57ab9156dd473d4ca4be9445c565f5a95a54e511..62a90c89cfc5b2cb32567ef7a8ef9a6c1cc30185

nataliasitko (approved):

triffer (commented):
The [AnnotationChangedPredicate](https://github.com/kyma-project/istio/blob/6b80b1464fec5ea6a8d2112bb2b6587d0b7d006a/controllers/istio_controller.go#L240) was introduced the ExternalName, but I think it's not harmful if we keep it for now.

We might also want to remove https://github.com/kyma-project/istio/blob/main/docs/user/migration/external-name-alias.md and https://github.com/kyma-project/istio/blob/6b80b1464fec5ea6a8d2112bb2b6587d0b7d006a/docs/user/_sidebar.md?plain=1#L9

kolodziejczak:
> The [AnnotationChangedPredicate](https://github.com/kyma-project/istio/blob/6b80b1464fec5ea6a8d2112bb2b6587d0b7d006a/controllers/istio_controller.go#L240) was introduced the ExternalName, but I think it's not harmful if we keep it for now.
> 
> We might also want to remove https://github.com/kyma-project/istio/blob/main/docs/user/migration/external-name-alias.md and
> 
> https://github.com/kyma-project/istio/blob/6b80b1464fec5ea6a8d2112bb2b6587d0b7d006a/docs/user/_sidebar.md?plain=1#L9

Let's keep the migration guide for now, since people usually are looking on main for the documentation - I'll create a followup for this. Also let's keep the predicate, since it's not harmful, and in the future it might be useful as we introduce another annotation that required the reconciliation.

triffer (approved):

PR #1117, opened 2024-11-13
ADR for extending Istio CR with support for AWS LB type

videlov:
**Description**

- ADR for extending Istio CR with support for AWS LB type

**Pre-Merge Checklist**

**Related issues**

### DIFF
--- a/docs/contributor/adr/0010-aws-nlb-support.md
+++ b/docs/contributor/adr/0010-aws-nlb-support.md
@@ -6,9 +6,9 @@ Proposed
 
 ## Context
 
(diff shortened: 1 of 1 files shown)


docs/contributor/adr/0010-aws-nlb-support.md
```diff
@@ -0,0 +1,61 @@
...
+kind: Istio
+metadata:
+  name: default
+  namespace: kyma-system
+spec:
+  config:
+    awsLoadBalancerType: nlb
+```
+
+* When switching to NLB Istio Module operator should apply additionally the following `serviceAnnotations` with Istio Operator Manifest:
```
triffer, line 47: What about this warning in the AWS load balancer controller docs:

> It's not recommended to modify or add the service.beta.kubernetes.io/aws-load-balancer-type annotation on an existing Service resource. If a change is desired, delete the existing Service resource and create a new one instead of modifying an existing Service.
>
>If you modify this annotation on an existing Service resource, you might end up with leaked LBC resources.

see https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.10/guide/service/nlb/

videlov: lets discuss on Monday, how we are going to approach this. Since downtime is a thing to consider.

Issue #1069, opened 2024-10-09
Clean up old Workflow runs


triffer:
**Description**

Clean up workflows that no longer exist, but are still visible in the [Actions UI](https://github.com/kyma-project/istio/actions). Those workflows can be easily detected by checking if they have a link to existing yaml file.
To avoid additional work this issue might be only done after the issue [Introduce naming scheme for workflow names and workflow files](https://github.com/kyma-project/istio/issues/1066) is completed, because in the mentioned issue the structure of workflows might change.

**Motivation**

Better visibility and up to date Actions on the UI.

**ToDo [Developer]**
- [x] Remove workflow runs that no longer exists in the pipelines

**PRs**
- [ ] https://github.com/kyma-project/istio/pull/1119

**AC [PO]**
- [x] Workflows that no longer exists are no longer visible in the [Actions UI](https://github.com/kyma-project/istio/actions)
- [x] Follow up for API Gateway is created or Workflows are also clean up for API Gateway

Ressetkk:
For future reference. The script lists 500 recent workflow runs and concurrently will issue a deletion 15 at a time.
```
gh run list --json databaseId -q '.[].databaseId' -w {WORKFLOW_FILE} --limit "500" |
            xargs -IID -P 15 gh run delete ID
```

PR #1119, opened 2024-11-14
feat: Add job that removes old workflow runs

Ressetkk:
/kind feat
/area ci

Add script and schedule workflow that removes workflow runs older than specific threshold.

Workflow runs every night at 0:11 UTC and removes up to 500 runs at a time.

#1069


triffer (approved):

triffer:
I don't know if we need to keep test execution jobs on the release branches longer than 60 days for some compliance reasons. But this should be already documented as part of the microdelivery. 🤔

Issue #925, opened 2024-07-08
[Automation] Create PR to update sec-scanners-config.yaml


barchw:
**Description**
Create a scheduled, weekly job, that will update the image version in `sec-scanners-config.yaml`

**Reasons**
- Automation of manual task

Pre-reqs:
- [x] https://github.com/kyma-project/api-gateway/issues/1234

**ToDos [Developer]**
- [x] Check with kyma-security if we actually can create an automated PR (we can use some goat bot like @kyma-gopher-bot)
- [x] Use Github service account creator tool to create this account

PR #1130, opened 2024-11-19
bug: fix fetching main image from registy

Ressetkk:
/kind bug
/area ci

There is a problem with properly assuming which tag should ba taken as the latest, because docker registry returns all tags for image only in alphabetical order. SHAs are not alphabetical, and sorting them is tedious task.

Because we build images and tag them with full commit SHA, instead of trying to fetch and sort all tags, grab latest HEAD of git repo and check if the tag exists on the repo. If it exists, update config. Otherwise, skip updating config update

#925

### DIFF
--- a/.github/workflows/update-sec-scanner.yaml
+++ b/.github/workflows/update-sec-scanner.yaml
@@ -5,26 +5,27 @@ on:
 permissions:
   contents: write
   pull-requests: write
+
 jobs:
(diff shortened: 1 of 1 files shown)


triffer (approved):
We don't fail when we are not able to fetch the latest image tag.
That can lead to a situation where if something is broking with fetching the image-tag, it will fail silently.

PR #1145, opened 2024-11-25
feat: workflow to add dependabot PRs to project

Ressetkk:
/kind feature
/area ci

Run workflow on pull request to automate adding dependabot PRs to board.


triffer (approved):
It's really shocking that we have to use those custom GraphQL requests to get this working. Personally, I think it would be better not to do all this just to add PRs to a project. 😄

PR #1161, opened 2024-12-03
Feat/release workflow adjustment

Ressetkk:
/kind feature
/area ci

Remove running main-integration.yaml workflow runs from release branches

Add separate workflow_call YAML call-integration-tests.yaml that runs during create-release.yaml workflow.

Add notification to Slack channel about new release

Rename pull-unit-lint.yaml to call-unit-lint.yaml

Run the following tests in the release workflow:
- units and lint
- istio upgrade integration test
- istio integration test (Gardener AWS/GCP and K3d)

#1163


.github/workflows/create-release.yaml
```diff
@@ -62,10 +62,20 @@ jobs:
         GO_BUILD_TAGS=experimental
       tags: "${{ github.event.inputs.name }}-experimental"
 
+  unit-tests:
+    uses: ./.github/workflows/call-unit-lint.yaml
+
+  integration-tests:
```
triffer, line 68: I'm just wondering if we also want to additionally test the experimental image? 🤔

triffer (approved):

.github/workflows/create-release.yaml
triffer: In daily it was agreed on that we keep it as it is for now.

PR #1238, opened 2025-01-09
Fetch depth on lint

barchw:
**Description**

- Fetch depth on lint

**Pre-Merge Checklist**

**Related issues**

### DIFF
--- a/.github/workflows/call-unit-lint.yaml
+++ b/.github/workflows/call-unit-lint.yaml
@@ -13,6 +13,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/call-unit-lint.yaml
```diff
@@ -13,6 +13,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          fetch-depth: 0
```
kolodziejczak, line 18: it will break release workflow, since there is no pull_request event

kolodziejczak: https://github.com/kyma-project/istio/pull/1237 is fixing it

PR #1176, opened 2024-12-06
Testing strategy for Istio module

barchw:
**Description**

Changes proposed in this pull request:

- Testing strategy for Istio module

**Pre-Merge Checklist**

- [ ] As a PR reviewer, verify code coverage and evaluate if it is acceptable.

**Related issues**

PR #1267, opened 2025-01-28
Adjust post workflow to the new test strategy

kolodziejczak:
**Description**

- Adjust post workflow to the new test strategy #1176 
- Introduce new version of the scripts related to Gardener e2e tests, similar to APIGateway module


hack/ci/scripts-v2/shoot_aws.yaml
```diff
@@ -0,0 +1,70 @@
+apiVersion: core.gardener.cloud/v1beta1
+kind: Shoot
+metadata:
+  name: ${CLUSTER_NAME}
```
mluk-sap, line 4: this doesn't contain 'pipeline link' that we've added recently in the API Gateway:
https://github.com/kyma-project/api-gateway/blob/7c76114e24bee93ffedc8e2773f620b9511cdd81/hack/ci/shoot_aws.yaml#L6

hack/ci/scripts-v2/shoot_gcp.yaml
```diff
@@ -0,0 +1,63 @@
+apiVersion: core.gardener.cloud/v1beta1
+kind: Shoot
+metadata:
+  name: ${CLUSTER_NAME}
```
mluk-sap, line 4: This doesn't contain the 'pipeline link' that we've added recently in the API Gateway:
https://github.com/kyma-project/api-gateway/blob/7c76114e24bee93ffedc8e2773f620b9511cdd81/hack/ci/shoot_gcp.yaml#L6

.github/workflows/post-main-workflow.yaml
```diff
@@ -0,0 +1,196 @@
...
+      context: .
+      build-args: |
+        VERSION=${{ github.ref_name }}-experimental
+        GO_BUILD_TAGS=experimental
+      tags: "${{ github.sha }}-experimental"
+
+  istio-upgrade-test-k3d:
+    name: Istio upgrade integration test K3D
+    runs-on: ubuntu-latest
+    needs: [ build-image ]
```
mluk-sap, line 56: Just wondering whether we should also introduce task like get-sha in the API Gateway module:
https://github.com/kyma-project/api-gateway/blob/7c76114e24bee93ffedc8e2773f620b9511cdd81/.github/workflows/post-main-integration.yaml#L30
In this case tests depends on get-sha, which depends on the build
It is more flexible, because the logic of getting proper sha is encapsulated in a separate task. It is then more extensible (changed in a single place).

mluk-sap (changes requested):

hack/ci/scripts-v2/provision-gardener.sh
```diff
@@ -0,0 +1,123 @@
...
+    echo "Could not apply shoot spec after 3 tries, exiting"
+    exit 1
+  fi
+  echo "Failed, retrying in 15s"
+  sleep 15
+done
+echo "Shoot template applied"
+
+echo "Waiting for cluster to be ready..."
+kubectl wait  --kubeconfig "${GARDENER_KUBECONFIG}" --for=condition=EveryNodeReady shoot/${CLUSTER_NAME} --timeout=25m
```
mluk-sap, line 97: We changed that in the API Gateway module recently, so we now wait 30 minutes for the 'status.lastOperation.state' to be Succeeded, see:
https://github.com/kyma-project/api-gateway/pull/1616/files#diff-6249dd23abca33d6832fc1726787aecd8935a32a338d1a79a6a142a62f0a37d4R98

.github/workflows/post-main-workflow.yaml
kolodziejczak: I think since this workflow doesn't run on `schedule`, we can avoid creating such task. I don't see any benefits of bringing it here, but I might be wrong. In the current state the image will be always tagged with the commit sha on which it was run and built simultaneously with the tests run, therefore it doesn't need to find the sha of the latest commit on which the image was built.

hack/ci/scripts-v2/provision-gardener.sh
kolodziejczak: Thanks, let me increase it here too

hack/ci/scripts-v2/shoot_aws.yaml
kolodziejczak: Thanks for catching it! I'll add it

hack/ci/scripts-v2/shoot_gcp.yaml
kolodziejczak: Thanks for catching it! I'll add it

.github/workflows/post-main-workflow.yaml
mluk-sap: We discussed it and the conclusion is that it is okay and we should rather adapt the API Gateway, because we split the main and schedule workflows there and the main one doesn't require get-sha.

mluk-sap (approved):

PR #1274, opened 2025-02-03
Fix release workflow yq

kolodziejczak:
**Description**

- Fix yq in the create release workflow

### DIFF
--- a/.github/workflows/release-create-release.yaml
+++ b/.github/workflows/release-create-release.yaml
@@ -44,6 +44,7 @@ jobs:
         shell: bash
         run: |
           if [[ $( ./yq eval ".protecode[0]" sec-scanners-config.yaml ) == "europe-docker.pkg.dev/kyma-project/prod/istio/releases/istio-manager:${{ github.event.inputs.version }}" ]]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/release-create-release.yaml
```diff
@@ -24,15 +24,27 @@ jobs:
...
+        env:
+          VERSION: "v4.45.1"
+          BINARY: "yq_linux_amd64"
+          EXPECTED_CHECKSUM: "654d2943ca1d3be2024089eb4f270f4070f491a0610481d128509b2834870049"
         run: |
-          sudo add-apt-repository ppa:rmescandon/yq
-          sudo apt update
-          sudo apt install yq -y
-
+          wget https://github.com/mikefarah/yq/releases/download/${VERSION}/${BINARY} -O ./yq && chmod +x ./yq
```
barchw, line 32: If changing, probably better to keep same as in makefile
https://github.com/kyma-project/api-gateway/blob/d0021a3fc118a116315398e8bb4458233fe82ab0/Makefile#L234

barchw (changes requested):

Ressetkk (approved):

PR #1388, opened 2025-04-07
Tests: configure TLS in gateway, use at least 3 machines in AWS clusters

mluk-sap:
**Description**

This PR should stabilize tests in AWS. At the moment tests sometimes resolve a non-working IP address of a DNS-based Load Balancer, which points to availability zone without any hosts, because the cluster is too small (2 machines vs 3 availability zones). This happens when health checks on AWS side are red or when they are being initialized.

- Added functions that generate dummy SSL certificate in tests
- Enabled HTTPs port in the Gateway installed by tests (otherwise AWS health checks in Load Balancer are red)
- Set cluster minimum size to 3 machines in AWS case, so there is at least one machine in every availability zone.

**Pre-Merge Checklist**

**Related issues**


tests/integration/steps/certificate.go
```diff
@@ -0,0 +1,108 @@
...
+		},
+		Data: map[string][]byte{
+			tlsKeyName: prvPEMBytes,
+			tlsCrtName: crtPEMBytes,
+		},
+		Type: corev1.SecretTypeTLS,
+	}
+
+	err = retry.Do(func() error {
+		err := k8sClient.Create(context.TODO(), secret)
```
kubicar, line 52: It should be not `context.TODO` (I know that is common practice in this repo that we don't set it in proper way for tests scenarios) IMHO we should use test context with timeout.

mluk-sap: Good point, I'll remember about it in future :)

PR #1428, opened 2025-05-15
bug: add missing gardener IP ranges for services

Ressetkk:
Gardener topology is a little bit different than standard Kubernetes one. DNS Service is working in host network mode. This disrupts the behaviour of NetworkPolicy.

This change requires the NetworkPolicy to contain the IP CIDR of services used for connectivity.

`100.104.0.0/13` is a default Service IP range when gardener cluster is created.

Tested against Gardeners on different providers, and locally in kind.


kubicar (approved):
It's not ideal that we need to implement it this way, but it's acceptable for test infrastructure purposes.

PR #1473, opened 2025-06-10
E2e framework

barchw:
**Description**

- ...

**Pre-Merge Checklist**

**Related issues**

### DIFF
--- a/tests/e2e/e2e/helpers/infrastructure/pod/pod.go
+++ b/tests/e2e/e2e/helpers/infrastructure/pod/pod.go
@@ -15,7 +15,7 @@ func CreatePod(t *testing.T, k8sClient client.Client, pod *v1.Pod) error {
 		t.Logf("Cleaning Pod: name: %s, namespace: %s", pod.Name, pod.Namespace)
 		err := k8sClient.Delete(setup.GetCleanupContext(), pod)
 		if err != nil {
-			t.Logf("Failed to delete pod: name: %s, namespace: %s, because: %w", pod.Namespace, pod.Name, err)
+			t.Logf("Failed to delete pod: name: %s, namespace: %s, because: %s", pod.Namespace, pod.Name, err)
 		}
 	})
 
--- a/tests/e2e/e2e/setup/namespace.go
+++ b/tests/e2e/e2e/setup/namespace.go
@@ -1,7 +1,6 @@
 package setup
 
 import (
-	"fmt"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"testing"
 
@@ -12,7 +11,7 @@ import (
 func CreateNamespaceForTest(t *testing.T, k8sClient client.Client, name string) {
 	namespace := &corev1.Namespace{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: fmt.Sprintf(name),
+			Name: name,
 		},
 	}
 


Ressetkk (commented):
Overall the idea on how to define the re-usable steps is pretty neat and I like it.
However, I still do not see it as a "framework". There is no clear and standarised convention on how to define steps inside a test case. For example in `tests/e2e/e2e/tests/createnamespacewithkubectl/create_ns_test.go` you define steps using a direct structure, which is fine, but in the `tests/e2e/e2e/tests/installistio/install_istio_test.go` the steps are hidden behind custom function `Steps()`. This lack of convention will confuse users and will generate the uncertainties even more. Also Not having assertions for example inside the step already defeats the purpose of logging, as you'd not be able to log the actions outside the executor.

I would see the executor as an implementation that wraps around all calls and actions inside the test, not only for specific subset of actions. For example each testcase (`TestFunction()` or `t.Run()`) equals separate executor that wraps the functions in test.

Moreover, in the examples you described assertions are not covered in the standarized log collection. This may be an overreaction on my side, but that behaviour would encourage developers to move outside the defined specification. That will make the entire framework useless for us.

I also miss the high-level discussions, diagrams, or just simple explanation how to write proper E2E tests withthis approach. It's supposed to be for E2E test, that means before the final go-to we still need to resolve uncertainties on how the initial cluster configuration would look like, how to set up and teardown infra between each test, so they do not break each other. I understand it will be a future step, but unless we figure this out, we can't rewrite existing tests.

tests/e2e/e2e/tests/resource/resource_test.go
joanbod, line 1: I like the way this test is grouped in the structure of: given, when, then. I think it would be more redable if other test files would be also grouped in this structure.

tests/e2e/e2e/tests/createpod/create_pod_test.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/kyma-project/istio/operator/tests/e2e/e2e/executor"
+	e2ePod "github.com/kyma-project/istio/operator/tests/e2e/e2e/steps/infrastructure/pod"
+)
+
+func TestPodCreation(t *testing.T) {
+	t.Parallel()
```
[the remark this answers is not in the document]
barchw: There's no reason not to run the tests parallel with t.Run in case we would want to
Probably this t.Parallel() at the the top does not make sense though 👍 
```suggestion
```

tests/e2e/e2e/executor/executor.go
```diff
@@ -0,0 +1,115 @@
...
+
+const (
+	ErrorPrefix   = "[ERROR] "
+	InfoPrefix    = "[INFO] "
+	TracePrefix   = "[TRACE] "
+	UntracePrefix = "[UNTRACE] "
+	DebugPrefix   = "[DEBUG] "
+)
+
+func Errorf(t *testing.T, template string, args ...interface{}) {
```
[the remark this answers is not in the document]
barchw: About this one I'm not sure, I actually considered moving the logging entirely out of the executor package.
There is no dependency to the executor besides to the `t`, so I would not say that this change is warranted. Keeping it separate allows calling the logger without instantiate executor (or in the steps that are called by the executor)
What I think could work instead is the logger being a seperate structure enveloping `t`

tests/e2e/e2e/executor/executor.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/stretchr/testify/assert"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/kyma-project/istio/operator/tests/e2e/e2e/setup"
+)
+
+type Step interface {
+	Description() string
+	Execute(*testing.T, client.Client) error
+	Cleanup(*testing.T, client.Client) error
```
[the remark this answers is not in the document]
kolodziejczak: Agree with @Ressetkk. It's also more idiomatic, for example like with the io.Reader and io.Writer in the std lib.

kolodziejczak (commented):
It looks very promising. I like it. Could we maybe just increase comments coverage? I think it would make much easier to onboard into it later

tests/e2e/e2e/executor/executor.go
```diff
@@ -0,0 +1,143 @@
...
+	// This is used for tracing of the test execution.
+	Description() string
+	// Execute runs the step, performing the necessary actions.
+	// It takes a testing.T instance for logging and a client.Client for Kubernetes operations.
+	Execute(*testing.T, client.Client) error
+}
+
+// Cleanuper is an interface that defines a cleanup operation.
+// A step should implement this interface if it requires cleanup after execution.
+type Cleanuper interface {
```
Ressetkk, line 27:
```suggestion
type Cleaner interface {
```

tests/e2e/e2e/executor/executor.go
```diff
@@ -0,0 +1,143 @@
...
+	if e.IsCi && !e.OnlyCleanup {
+		logging.Infof(e.t, "Skipping cleanup in CI environment")
+		return
+	}
+
+	logging.Tracef(e.t, "Starting cleanup")
+	defer logging.Untracef(e.t, "Finished cleanup")
+
+	// Perform cleanup in reverse order
+	for step := e.steps.Pop(); step != nil; step = e.steps.Pop() {
```
Ressetkk, line 131: Implement `e.steps.Empty()` method and use it as below. See: https://go.dev/play/p/9FXwxp5n4-h
```go
for !e.steps.Empty() {
```

Ressetkk (commented):
2 more CR comments

tests/e2e/e2e/executor/executor.go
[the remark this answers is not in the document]
barchw: My general point of view is more towards @mluk-sap comment, as having to always implement that func means that it will also always come up in the Pull Request review. So the reviewer sees that the cleanup func is not doing anything on purpose, not just because the implementer forgot to implement the cleanup.

tests/e2e/e2e/steps/exec/bash.go
```diff
@@ -0,0 +1,68 @@
...
+	CleanupCmd string
+}
+
+func (c *Command) Description() string {
+	return fmt.Sprintf("Executing command: %s", c.Command)
+}
+
+func (c *Command) Execute(t *testing.T, _ client.Client) error {
+	splitCommand := strings.Split(c.Command, " ")
+	cmd := exec.Command(splitCommand[0], splitCommand[1:]...) // #nosec G204
```
mluk-sap, line 30: It would be better to provide commands as arrays and avoid split logic here, which would be more 'correct' and expected by the `exec.Command`, correctly handling parameters with spaces, etc.

tests/e2e/e2e/logging/logging.go
```diff
@@ -0,0 +1,31 @@
...
+func Errorf(t *testing.T, template string, args ...interface{}) {
+	t.Logf(ErrorPrefix+template, args...)
+}
+
+func Infof(t *testing.T, template string, args ...interface{}) {
+	t.Logf(InfoPrefix+template, args...)
+}
+
+func Debugf(t *testing.T, template string, args ...interface{}) {
+	t.Logf(DebugPrefix+template, args...)
```
mluk-sap, line 22: It would be nice to conditionally invoke it depending on the current log level.

tests/e2e/e2e/logging/logging.go
```diff
@@ -0,0 +1,31 @@
...
+}
+
+func Tracef(t *testing.T, template string, args ...interface{}) {
+	t.Logf(TracePrefix+template, args...)
```
mluk-sap, line 26: It would be nice to conditionally invoke it depending on the current log level.

tests/e2e/e2e/steps/installistio/install_istio.go
```diff
@@ -0,0 +1,29 @@
+package installistio
+
+import (
+	"github.com/kyma-project/istio/operator/tests/e2e/e2e/executor"
+	"github.com/kyma-project/istio/operator/tests/e2e/e2e/steps/exec"
+)
+
+func Steps() []executor.Step {
+	createNamespaceStep := &exec.Command{
+		Command:    "kubectl create namespace kyma-system",
```
mluk-sap, line 10: See above comment about using arrays in commands.

tests/e2e/e2e/tests/createnamespacewithkubectl/create_ns_test.go
```diff
@@ -0,0 +1,46 @@
...
+		// Setup Infra
+		testExecutor := executor.NewExecutorWithOptionsFromEnv(t)
+		defer testExecutor.Cleanup()
+
+		createNs := &bashStep.Command{
+			Command:    "kubectl create namespace test-namespace",
+			CleanupCmd: "kubectl delete namespace test-namespace",
+		}
+
+		err := testExecutor.RunStep(createNs)
```
mluk-sap, line 27: Loudly thinking:
The idea of having steps and interfaces here causes issues:
- it os not very readable
- tests would have mainly testExecutor.RunStep invocations
- we need to grab output, which is also against generic 'RunStep' here
Let's discuss it later...

tests/e2e/e2e/executor/executor.go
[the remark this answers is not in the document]
Ressetkk: > Another point of view: the advantage of having a single interface is that it forces the end-developer to always **think** about the cleanup.

But if the step is not considered as stateful, where there is no cleanup actions on an environment, then why would developer have to implement empty function just to conform the interface? That would go against the concept of simplicity in this case.

tests/e2e/e2e/tests/createnamespacewithkubectl/create_ns_test.go
barchw: after reviewing our options we decided to instead implement the helpers as a pure go funcs (without interfaces)

tests/e2e/e2e/tests/createnamespacewithkubectl/create_ns_test.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+func TestCreateNsWithKubectl(t *testing.T) {
+	t.Parallel()
+
+	t.Run("Create Namespace", func(t *testing.T) {
+		t.Parallel()
+		// Setup Infra
+		testExecutor := executor.NewExecutorWithOptionsFromEnv(t)
+		defer testExecutor.Cleanup()
```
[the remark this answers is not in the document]
barchw: We will use t.Cleanup() since we are removing the executor and step interface

PR #1497, opened 2025-06-30
Setup for support of arm64 image building

barchw:
**Description**

- adapt Dockerfile
- adapt workflow of image builder
- adapt make target

**Pre-Merge Checklist**

**Related issues**

### DIFF
--- a/.github/workflows/post-main-workflow.yaml
+++ b/.github/workflows/post-main-workflow.yaml
@@ -50,6 +50,7 @@ jobs:
         VERSION=${{ github.ref_name }}-experimental
         GO_BUILD_TAGS=experimental
       tags: "${{ github.sha }}-experimental"
+      build-engine: buildx
 
   istio-upgrade-test-k3d:
     name: Istio upgrade integration test K3D
--- a/.github/workflows/post-main-workflow.yaml
+++ b/.github/workflows/post-main-workflow.yaml
@@ -36,6 +36,7 @@ jobs:
       build-args: |
         VERSION=${{ github.ref_name }}
       tags: "${{ github.sha }}"
+      build-engine: buildx
 
   build-experimental-image:
     name: Build manager image - experimental
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM golang:1.24.4-alpine AS builder
+FROM --platform=$BUILDPLATFORM golang:1.24.4-alpine AS builder
 ARG TARGETOS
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -103,13 +103,16 @@ build: generate fmt vet ## Build manager binary.
 run: manifests install build create-kyma-system-ns ## Run a controller from your host.
 	ISTIO_INSTALL_BIN_PATH=$(ISTIO_INSTALL_BIN_PATH) go run ./main.go
 
+TARGET_OS ?= linux
```
kolodziejczak, line 106: We have defaults also set in Dockerfile. I think we should pick only one place to do that

.github/workflows/post-main-workflow.yaml
```diff
@@ -36,6 +36,7 @@ jobs:
       build-args: |
         VERSION=${{ github.ref_name }}
       tags: "${{ github.sha }}"
+      build-engine: buildx
```
kolodziejczak, line 39: It's the default value from today. I don't mind it being explicitly set, but then we should adjust experimental build in this workflow, and also builds in the release workflow.

Dockerfile
```diff
@@ -31,7 +31,7 @@ RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH:-amd64} go build -
 
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:nonroot
+FROM gcr.io/distroless/static-debian12:nonroot
```
kolodziejczak, line 34: can I ask for the reason for this change?

kolodziejczak (changes requested):
We need to adjust also experimental build in post main workflow and both build on the release

.github/workflows/post-main-workflow.yaml
kolodziejczak: sure, please add it in the rest of the build jobs too

kolodziejczak:
The question, what do we do about https://github.com/kyma-project/istio/blob/main/.github/actions/build-manager-image/action.yaml?

barchw:
> The question, what do we do about https://github.com/kyma-project/istio/blob/main/.github/actions/build-manager-image/action.yaml?

We do not depend on the arm64 building anywhere in the project, so I think we should leave it as it is. The image build there is not for production use.

kolodziejczak (approved):

PR #1510, opened 2025-07-08
poc: handling test resources with testdata and helpers

Ressetkk:
POC of handling YAML resources in E2E tests in go-native tests
Used egress e2e test as real-life example

add oauth2-mock helper

extra: poc of using e2e-framework for setting up entire cluster suite


tests/e2e/egress/egress_test.go
barchw, line 1: Please move the helper function `setupIstio` along with `teardownIstio` to the `tests/e2e/e2e/helpers` pkg.

Might also be worth do consider making `curlInPod` another helper.

tests/e2e/pkg/helpers/ns/namespace.go
barchw, line 1: Please move this under `tests/e2e/e2e/helpers` for later reuse.

tests/e2e/egress/egress_test.go
```diff
@@ -57,400 +48,132 @@ func TestE2EEgressConnectivity(t *testing.T) {
...
+			require.NoError(t, ns.CreateNamespace(ctx, t, runID, cfg,
+				ns.WithLabels(map[string]string{"istio-injection": "enabled"}),
+			))
 			if tt.applyEgressConfig {
-				err := applyEgressConfig(ctx, c, ns)
-				if err != nil {
-					t.Errorf("failed to create egress resources: %v", err)
-				}
+				t.Log("Applying egress config for the test: ", runID)
+				require.NoError(t, decoder.DecodeEachFile(ctx, fsys, "egress_config.yaml", decoder.CreateHandler(r), decoder.MutateNamespace(runID)))
```
barchw, line 93: We should make sure that we are using the `decoder` package, or the `tests/e2e/e2e/helpers/infrastructure/yamlfile` everywhere.

tests/e2e/egress/egress_test.go
```diff
@@ -57,400 +48,132 @@ func TestE2EEgressConnectivity(t *testing.T) {
...
 			if err != nil && !tt.expectError {
 				t.Errorf("got an error but shouldn't have: %v", err)
 			}
 			if err == nil && tt.expectError {
 				t.Error("didn't get an error but expected one")
 			}
 
-			t.Log(string(out))
+			t.Log(stdout.String())
+			t.Log(stderr.String())
```
barchw, line 116: Those logs should have some prefix / explanation of what they are logging.
e.g.:

```go
t.Logf("Exec in pod stderr: %s, stderr.String())
```

tests/e2e/pkg/helpers/ns/namespace.go
```diff
@@ -0,0 +1,51 @@
...
+type CreateOpts func(object k8s.Object)
+
+func WithLabels(labels map[string]string) CreateOpts {
+	return func(object k8s.Object) {
+		object.SetLabels(labels)
+	}
+}
+
+func CreateNamespace(ctx context.Context, t *testing.T, name string, cfg *envconf.Config, opts ...CreateOpts) error {
+	r, err := resources.New(cfg.Client().RESTConfig())
```
barchw, line 25: Is there an option to use the k8s client with logging?
`tests/e2e/e2e/setup/client.go`

barchw (changes requested):

tests/e2e/egress/egress_test.go
```diff
@@ -1,49 +1,40 @@
...
 	ctx := context.Background()
-	cfg := config.GetConfigOrDie()
-	c, err := initK8sClient(cfg)
-	if err != nil {
-		t.Fatalf("failed to init cluster client: %v", err)
-	}
-	setupIstio(ctx, c, t)
+	path := conf.ResolveKubeConfigFile()
+	cfg := envconf.NewWithKubeConfig(path)
+	fsys := os.DirFS("testdata")
```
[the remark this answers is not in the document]
Ressetkk: No. `tests/e2e/e2e/setup/client.go` uses underlying client interface from controller-runtime. It only supports CRUD operations. The resources client applies also non-CRUD operations ootb, like exec commands in a pod.

This was also an issue I wanted to tackle in the tests somehow. This implementation resolves this problem.

tests/e2e/pkg/helpers/ns/namespace.go
Ressetkk: Looks like so.
it uses the same rest.Config struct for instantiating client.

tests/e2e/pkg/helpers/ns/namespace.go
Ressetkk: You should keep helpers in pkg. Otherwise, the helpers package in `tests/e2e/e2e/helpers` should be consolidated into `tests/pkg`. I would avoid going against the standard structure, at least for reusable code.

barchw: That's also ok if you move them under `pkg`. Just wanted to make sure that those are in the same place.

tests/e2e/egress/egress_test.go
```diff
@@ -57,400 +48,132 @@ func TestE2EEgressConnectivity(t *testing.T) {
...
-			t.Log(string(out))
+			t.Log(stdout.String())
+			t.Log(stderr.String())
 		})
 	}
 }
 
-func setupIstio(ctx context.Context, c client.Client, t *testing.T) {
-	// this function is a helper
+func setupIstio(ctx context.Context, fsys fs.FS, cfg *envconf.Config, t *testing.T) error {
```
barchw, line 121: This function will not be possible to be reused in the current form.
This is caused by:
- It uses a local `istio_customresource.yaml` instead of the release artifact `https://github.com/kyma-project/istio/releases/latest/download/istio_customresource.yaml`
- It does not create the istio manager resources

Additionally it has a dependency to structured `operatorv1alpha2.Istio` API, which will become a unnecessary dependency if we were to use it in API Gateway

barchw: Another thing is that it probably should use `make deploy` (or something equivalent) in case of installing the module from the same repo (so installing Istio here, and APIGateway if such a helper would exist for APIGateway) as we have the need to test on the current image.

tests/e2e/pkg/helpers/ns/namespace.go
```diff
@@ -0,0 +1,51 @@
...
+
+type CreateOpts func(object k8s.Object)
+
+func WithLabels(labels map[string]string) CreateOpts {
+	return func(object k8s.Object) {
+		object.SetLabels(labels)
+	}
+}
+
+func CreateNamespace(ctx context.Context, t *testing.T, name string, cfg *envconf.Config, opts ...CreateOpts) error {
```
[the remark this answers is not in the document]
Ressetkk: > Generally I would expect that the `*testing.T` will be the first parameter, then `ctx`, then else. (There are other occurrences of the same)

According to the context godoc `ctx` should be first argument. https://pkg.go.dev/context
>The Context should be the first parameter, typically named ctx:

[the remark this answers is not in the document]
barchw: That being said it should definitely be consistent (so e.g. T is second), we have it different there:
```go
func setupIstio(ctx context.Context, fsys fs.FS, cfg *envconf.Config, t *testing.T) error 
```

tests/e2e/tests/oauth2mock/oauth2_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+		assert.Equal(t, "mock-oauth2-server.oauth2-mock.svc.cluster.local", m.IssuerURL)
+		assert.Equal(t, "oauth2-mock.local.kyma.dev", m.Subdomain)
+		assert.Equal(t, "https://oauth2-mock.local.kyma.dev/oauth2/token", m.TokenURL)
+	})
+
+	t.Run("deploying oauth2mock in a custom namespace", func(t *testing.T) {
+		t.Parallel()
+		m, err := oauth2mock.DeployMock(t, "local.kyma.dev",
+			oauth2mock.Options{Namespace: "custom-oauth2-mock"})
```
mluk-sap, line 30: Shouldn't we make this namespace unique? (append testId, etc.)
Then it is possible to rerun the test even if previous run polluted the cluster with a namespace (and was killed, etc.)

mluk-sap (dismissed):

Issue #1523, opened 2025-07-11
EnvoyFilter proxy protocol applied by module does not have labels updated during reconciliation


kolodziejczak:
**Description**

During `EnvoyFilter` `proxy-protocol` reconciliation controller does not update labels. Some time ago there was a new label introduced `kyma-project.io/module: istio` to the mentioned EF. For clusters that have the proxy protocol EF applied before the label was introduced, there won't be the label added. It creates 2 issues:
- old proxy protocol EF without label is detected as user created
- proxy protocol EF does not have mandatory label which allows to quickly filter resources that are owned by Kyma/Module

Moreover the logic is shared in a common resource manager. The issue most likely applies to any other resources that are reconciled by Istio Module.

**Expected result**

Reconciled resources are fully reconciled, including labels.

**Actual result**

Reconciled resources do not have labels updated during the reconciliation.

PR #1527, opened 2025-07-15
bug: Fix labels reconciliation in resource manager

kolodziejczak:
**Description**

- Fix labels reconciliation in resource manager - now it appends new labels if we introduce them in yaml definitions
- Fix `app.kubernetes.io/version` label which was broken due to build system issue - go build empty `-tags` argument caused `-ldflas` flag being treated as the argument for `-tags`
- introduced regression e2e test for `app.kubernetes.io/version` label

**Pre-Merge Checklist**

**Related issues**

#1523

### DIFF
--- a/tests/integration/scenario.go
+++ b/tests/integration/scenario.go
@@ -3,6 +3,8 @@ package integration
 import (
 	"github.com/cucumber/godog"
 	"github.com/kyma-project/istio/operator/tests/integration/steps"
+	"github.com/kyma-project/istio/operator/tests/integration/testsupport"
+	"log"
 )
 
 func initScenario(ctx *godog.ScenarioContext) {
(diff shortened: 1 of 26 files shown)


tests/integration/testsupport/environments.go
```diff
@@ -0,0 +1,60 @@
...
+type EnvVar struct {
+	Name    string
+	Default string
+}
+
+var requiredEnvs = []EnvVar{
+	{"OPERATOR_VERSION", "dev"},
+}
+
+func EnvironmentVariables() error {
```
kubicar, line 17: pls rename it for example to:  `LoadEnvironmentVariables`

kubicar (changes requested):

tests/integration/scenario.go
```diff
@@ -11,6 +13,12 @@ func initScenario(ctx *godog.ScenarioContext) {
 	ctx.After(istioCrTearDown)
 	t := steps.TemplatedIstioCr{}
 
+	err := testsupport.LoadEnvironmentVariables()
+	if err != nil {
+		panic("Failed to set environment variables: " + err.Error())
+	}
+	log.Printf("initializing tests with Istio Module version: %s", testsupport.GetOperatorVersion())
```
kubicar, line 20: log.Printf("initializing tests with Istio Module version: %s", testsupport.GetOperatorVersion())
should be in place where it is initialize.

kolodziejczak: yeah, that's a good point

nataliasitko (approved):

Issue #1621, opened 2025-09-26
Istio native sidecar enabled by default


strekm:
**Description**

Istio is going into direction to have native sidecar enabled by default. atm Istio module is configuring that option to false. Introduce grace period supporting classic sidecar in compatibility mode and enabling native sidecar by default. All workloads should be restarted to react to that setting.

**ACs:**
- [ ] native sidecars enabled by default
- [ ] all workloads restarted on setting change
- [ ] native sidecar setting can be reverted with compatibility mode
- [ ] relevant documentation updated

**TODO:**
- [ ] Remove ENABLE_NATIVE_SIDECAR env variable from IstioOperator yamls
- [ ] Add ENABLE_NATIVE_SIDECAR env to the compatibilityMode
- [ ] Ensure Istio bump patch or if doesn't happen implement restarted for the workloads based on the ENABLE_NATIVE_SIDECAR env var change
- [ ] update all related docs

PR #1632, opened 2025-10-03
Native sidecars documentation update

nataliasitko:
**Description**

- Update documentation on native sidecars and adjust it to the changes that Isito module 1.22 introduces

**Pre-Merge Checklist**

- [ ] The code coverage is acceptable.
- [ ] Release notes for the introduced changes are created.
- [ ] If Kubebuilder changes were made, you ran `make generate-manifests` and committed the changes before the merge.
- [ ] Pre-existing managed resources are correctly handled.
- [ ] There is no upgrade downtime.
- [ ] For infrastructure changes, you checked if the changes affect the hyperscaler's costs.
- [ ] RBAC settings are as restrictive as possible.
- [ ] If any new libraries are added, you verified license compliance and maintainability, and made a comment in the PR with details. We only allow stable releases to be included in the project.
- [ ] You checked if this change should be cherry-picked to active release branches.
- [ ] The configuration does not introduce any additional latency.

**Related issues**

https://github.com/kyma-project/istio/issues/1621

### DIFF
--- a/docs/user/00-20-istio-proxy-as-native-sidecar.md
+++ b/docs/user/00-20-istio-proxy-as-native-sidecar.md
@@ -97,10 +97,10 @@ The Istio module fully supports Istio proxies run as regular containers and as i
     test-job-2tt95   0/2     Completed         0          22s
     ```
 
-The key differences when compared to the `istio-proxy` run as a regular container:
-  - When the Pod starts, it has the `Init` status with two containers.
-  - The application container only starts after the `istio-proxy` native sidecar is fully operational.
-  - After the main container completes its job, the Pod finishes with the status `Completed`.
+    The key differences when compared to the `istio-proxy` run as a regular container:
+    - When the Pod starts, it has the `Init` status with two containers.
+    - The application container only starts after the `istio-proxy` native sidecar is fully operational.
+    - After the main container completes its job, the Pod finishes with the status `Completed`.
 
 4. List init containers:
 


kolodziejczak (dismissed):

docs/user/troubleshooting/03-45-jobs-cant-finish.md
```diff
@@ -5,16 +5,12 @@
 Pods created by Jobs remain stuck in the `NotReady` status after the main containers have finished.
 
 ## Cause
+By default, the Istio module injects `istio-proxy` containers as native sidecars. However, you can also set the annotation `sidecar.istio.io/nativeSidecar` annotation to `"false"` for a specific Pod. This annotation overwrites the default setting and indicates that instead of native sidecars, the annotated Pod must be injected with a regular sidecar container.
```
IwonaLanger, line 8:
```suggestion
By default, the Istio module injects `istio-proxy` containers as native sidecars. However, you can also set the annotation `sidecar.istio.io/nativeSidecar` to `"false"` for a specific Pod. This annotation overwrites the default setting and indicates that instead of native sidecars, the annotated Pod must be injected with a regular sidecar container.
```

docs/user/troubleshooting/03-46-init-containers-cant-access-network.md
```diff
@@ -5,16 +5,13 @@
 Init containers can't access the network.
 
 ## Cause
+If Istio injection is enabled, the `istio-proxy` container intercepts all network traffic from all containers. By default, the Istio module injects `istio-proxy` containers as native sidecars. However, you can also set the annotation `sidecar.istio.io/nativeSidecar` annotation to `"false"` for a specific Pod. This annotation overwrites the default setting and indicates that instead of native sidecars, the annotated Pod must be injected with a regular sidecar container.
```
IwonaLanger, line 8:
```suggestion
If Istio injection is enabled, the `istio-proxy` container intercepts all network traffic from all containers. By default, the Istio module injects `istio-proxy` containers as native sidecars. However, you can also set the annotation `sidecar.istio.io/nativeSidecar` to `"false"` for a specific Pod. This annotation overwrites the default setting and indicates that instead of native sidecars, the annotated Pod must be injected with a regular sidecar container.
```

IwonaLanger (approved):

docs/user/troubleshooting/03-46-init-containers-cant-access-network.md
```diff
@@ -5,16 +5,13 @@
...
+Init containers are started before regular containers. If `istio-proxy` is a regular sidecar, it doesn't work when init containers are running. As a result, init containers don't have network access.
 
 ## Solution
 
-Inject `istio-proxy` as a native sidecar container.
-
-To do this, set the `sidecar.istio.io/nativeSidecar` annotation in the Pod to `"true"`.
-
-See the following example:
+Check if your Pod or the Pod's template contains the annotation `sidecar.istio.io/nativeSidecar: "false"`. See the following example of an annotated Pod:
```
mluk-sap, line 14: There may be more reasons for having istio-proxy as regular container:
- old Istio module version
- compatiblityMode enabled
- container may not be restarted yet after the upgrade (restart is stuck for whatever reason, Istio module not working, etc.)
- some bug that we haven't discovered yet
 
So maybe we could write it in the simpler way, like:
1. Check whether the istio-proxy is declared as initContainer or container
2. If it is an initContainer -> it means that it already runs as native sidecar and the networking issues are caused by something else
3. If it is a regular container -> then it is NOT a native sidecar and you should switch it to native sidecar by an annotation sidecar.istio.io/nativeSidecar=true

This would be quite simple, because the troubleshooting algorithm comes down to a single check in point '1'.

PR #1672, opened 2025-10-31
Added IstioInstall function that doesn't use port forward

mluk-sap:
**Description**

- Added IstioInstall function that calls Istio installation in simplified way (when compared to original Istio impl):
  - hardcoded parameters like force, skip confirmation, etc.
  - removed detectIstioVersionDiff function call with all related calls (getting tags with Istio version), which was calling non-fips compliant port forwarding in order to get current Istio version from istiod tcp port 15014
  - removed obsolete verify=false parameter
  - removed applyFlagAliases call which always returns empty struct in our case
- Added validation for missing command line argument with iop file names
- Added `GODEBUG="fips140=only,tlsmlkem=0"` to Dockerfile, so the controller image is always fips compliant

**Pre-Merge Checklist**

- [x] The code coverage is acceptable.
- [x] Release notes for the introduced changes are created.
- [x] If Kubebuilder changes were made, you ran `make generate-manifests` and committed the changes before the merge.
- [x] Pre-existing managed resources are correctly handled.
- [x] There is no upgrade downtime.
- [x] For infrastructure changes, you checked if the changes affect the hyperscaler's costs.
- [x] RBAC settings are as restrictive as possible.
- [x] If any new libraries are added, you verified license compliance and maintainability, and made a comment in the PR with details. We only allow stable releases to be included in the project.
- [x] You checked if this change should be cherry-picked to active release branches.
- [x] The configuration does not introduce any additional latency.

**Related issues**

### DIFF
--- a/docs/release-notes/1.24.0.md
+++ b/docs/release-notes/1.24.0.md
@@ -1,8 +1,8 @@
 ## New Features
 
-- Istio controller image is now FIPS compliant, which is enforced by `fips140=only,tlsmlkem=0` flags.
+- Istio controller image is now running in FIPS140 compliancy mode, which is controlled by the `fips140=only,tlsmlkem=0` GODEBUG flags.
   We simplified the Istio installation, so it doesn't use port forwarding to get the Istio version.
-  It also doesn't print the confusing message anymore about the Istio downgrade.
+  It also doesn't print the message about the Istio downgrade.
   See [#1672](https://github.com/kyma-project/istio/pull/1672).
 
 ## Fixed Bugs
--- a/cmd/istio-install/main.go
+++ b/cmd/istio-install/main.go
@@ -64,10 +64,6 @@ func main() {
 }
 
 func IstioInstall(kubeClient kube.CLIClient, iopFileNames []string, l clog.Logger) error {
-	if err := k8sversion.IsK8VersionSupported(kubeClient, l); err != nil {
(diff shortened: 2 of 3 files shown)


docs/release-notes/1.24.0.md
```diff
@@ -0,0 +1,8 @@
+## New Features
+
+- Istio controller image is now FIPS compliant, which is enforced by `fips140=only,tlsmlkem=0` flags.
```
nataliasitko:
```suggestion
- Istio controller image is now FIPS compliant, which is enforced by the `fips140=only,tlsmlkem=0` flags.
```

barchw: I would avoid statements that we are not sure to be true ;)

```suggestion
- Istio controller image is now running in FIPS140 compliancy mode, which is controlled by `fips140=only,tlsmlkem=0` GODEBUG flags.
```

Dockerfile
```diff
@@ -36,6 +36,7 @@ WORKDIR /
 COPY --from=builder /istio-build/manager .
 COPY --from=builder /istio-build/istio_install .
 
+ENV GODEBUG="fips140=only,tlsmlkem=0"
```
barchw: I suggest that we add that environment variable to the chart as well even though it is not explicitly needed

barchw (changes requested):

barchw (approved):

PR #1889, opened 2026-02-24
Add documentation about trustDomain

barchw:
**Description**

- Add documentation explaining trust domain 

**Related issues**

### DIFF
--- a/docs/user/00-30-trust-domain.md
+++ b/docs/user/00-30-trust-domain.md
@@ -1,5 +1,5 @@
 # Istio Trust Domain
-Configuring trust domain is used to align identities across multiple clusters, enable seamless multi-mesh mTLS communication with shared root CAs, and match organizational naming standards.
+Configure a trust domain to align identities across multiple clusters, enable seamless multi-mesh mTLS communication with shared root CAs, and match organizational naming standards.
 
 ## What Is a Trust Domain?
 A trust domain defines the security boundary of a service mesh.


docs/user/technical-reference/05-30-trust-domain.md
```diff
@@ -0,0 +1,109 @@
...
+
+```
+spiffe://<trust-domain>/ns/<namespace>/sa/<service-account>
+```
+
+By default, Istio uses the **trustDomain** `cluster.local`. You can override it to align identities across clusters or to match organizational naming standards.
+
+## Configure trustDomain in the Istio custom resource
+
+Set the **trustDomain** under `spec.config.trustDomain` in the Istio CR. If omitted, the default value `cluster.local` is used.
```
nataliasitko:
```suggestion
Set the trust domain under **spec.config.trustDomain** in the Istio CR. If omitted, the default value `cluster.local` is used.
```

docs/user/technical-reference/05-30-trust-domain.md
```diff
@@ -0,0 +1,109 @@
...
+  config:
+    trustDomain: "example.com"
+```
+
+## Where does the **trustDomain** appear?
+
+Some of the key places where the **trustDomain** is used include:
+
+- **Istio proxy identity certificate**: The **trustDomain** is included in the SAN of the certificate issued to every istio proxy in the mesh.
+- **Authorization policies**: When defining AuthorizationPolicies, you can either use the default `cluster.local`
```
nataliasitko:
```suggestion
- AuthorizationPolicies: When defining AuthorizationPolicies, you can either use the default value `cluster.local`
```

docs/user/technical-reference/05-30-trust-domain.md
```diff
@@ -0,0 +1,109 @@
...
+or your custom **trustDomain** in the SPIFFE IDs to specify which workloads the policy applies to.
+
+## Multi-mesh seamless mTLS communication
```
nataliasitko:
```suggestion
## Multi-Mesh Seamless mTLS Communication
```

docs/user/technical-reference/05-30-trust-domain.md
```diff
@@ -0,0 +1,109 @@
...
+
+When communicating between multiple meshes, in case a [shared Root CA is plugged in](../00-25-plug-in-istio-ca.md) to issue proxy certificates in both meshes,
+Gateway with **tls.mode: ISTIO_MUTUAL** can be used to seamlessly allow mTLS communication between the two meshes without the need to set up `mTLS` on the Gateway manually.
+In this case, the **trustDomain** can be used to differentiate the source of the request in the upstream mesh (from inside the mesh vs from another mesh),
+as the SPIFFE ID in the client certificate will include the **trustDomain** of the originating mesh.
```
nataliasitko:
```suggestion
If a [shared Root CA is plugged in](../00-25-plug-in-istio-ca.md) to issue proxy certificates in multiple meshes, you can use a Gateway with **tls.mode: ISTIO_MUTUAL** to seamlessly allow mTLS communication between them. In this case, you don't need to set up `mTLS` on the Gateway manually.

Additionally, you can use the trust domain to differentiate the source of requests in the upstream mesh (whether they come from within the same mesh or from another mesh). This is possible because the SPIFFE ID in the client certificate includes the trust domain of the originating mesh.
```

docs/user/technical-reference/05-30-trust-domain.md
```diff
@@ -0,0 +1,109 @@
...
+
+In this case an example Gateway configuration in the target mesh can be as simple as:
```
nataliasitko:
```suggestion
In this case, you can use the following example Gateway configuration in the target mesh:
```

docs/user/00-30-trust-domain.md
```diff
@@ -0,0 +1,110 @@
...
+```
+
+## Migration Considerations
+
+Changing the trust domain is a sensitive operation. Once you update it, existing workloads may fail authentication until they receive new certificates.
+The default `cluster.local` value is treated as a special case that is compatible with custom trust domains.
+This means that setting the trust domain from `cluster.local` to a new, custom value doesn't cause authentication failures
+for existing workloads until they are restarted and receive new certificates with the updated trust domain.
+With that said, it is still recommended to plan the change carefully and perform it during a maintenance window to minimize disruption.
+For a detailed migration guide,
```
werdes72: Newline is not needed here.

docs/user/00-30-trust-domain.md
```diff
@@ -0,0 +1,110 @@
+# Configuring a Trust Domain
```
nataliasitko:
```suggestion
# Configure a Trust Domain
```

PR #1944, opened 2026-03-24
Sync Istio 1.28.5

barchw:
**Description**

- Sync istio images 1.28.5


fips-images.yaml
```diff
@@ -1,17 +1,17 @@
 images:
-  - source: "europe-docker.pkg.dev/kyma-project/restricted-prod/sap.com/istio-pilot-fips:1.28.3@sha256:f12d2542fad3ff4a3a61f1fa92af4d46522dea9f94ed86c2ae0d1038199bdac1"
```
kolodziejczak: This SHA has its pros and cons. I'd still leave it here, to make it explicit what was synced before. In other case to make sure that the images will differ during the update under the same tag, developer has to check manually what SHA is in the registry, or just believe the version differs.

PR #2024, opened 2026-05-04
Sync Istio 1.29.2 with registry

kolodziejczak:
**Description**

- Sync Istio 1.29.2 with registry


kubicar:
@kolodziejczak we also need to bump entries in fips-images.yaml file

kolodziejczak:
@kubicar I like to keep it in separate PRs, I'll do the next one after

kubicar (approved):

PR #2045, opened 2026-05-18
test:  add comprehensive tests for clusterconfig current behavior 

kubicar:
**Description**

- test: add comprehensive tests for clusterconfig current behavior
- move out from Ginkgo

**Pre-Merge Checklist**

- [ ] The code coverage is acceptable.
- [ ] Release notes for the introduced changes are created.
- [ ] If Kubebuilder changes were made, you ran `make manifests` and committed the changes before the merge.
- [ ] Pre-existing managed resources are correctly handled.
- [ ] There is no upgrade downtime.
- [ ] For infrastructure changes, you checked if the changes affect the hyperscaler's costs.
- [ ] RBAC settings are as restrictive as possible.
- [ ] If any new libraries are added, you verified license compliance and maintainability, and made a comment in the PR with details. We only allow stable releases to be included in the project.
- [ ] You checked if this change should be cherry-picked to active release branches.
- [ ] The configuration does not introduce any additional latency.
- [ ] You checked if Busola updates are needed.
- [ ] If you added a predicate for restarters, check if the **lastAppliedConfiguration** annotation is properly updated after the given restart is executed.

**Related issues**


internal/clusterconfig/clustersize_test.go
```diff
@@ -0,0 +1,95 @@
+package clusterconfig_test
```
kolodziejczak, line 1: nit: it's not a real world scenario probably / super corner case (eg. when the operator runs outside the cluster, and the cluster really has no nodes atm + control plane outside), but we could consider how the code behaves if the k8s client does not have any node to return. feel free to skip this, if you consider it irrelevant

internal/clusterconfig/loadbalancer_test.go
```diff
@@ -0,0 +1,246 @@
...
+	node := newOpenStackNode("openstack-node-1")
+	testClient := newTestClient(t, node)
+
+	// When
+	provider, err := clusterconfig.GetClusterProvider(context.Background(), testClient)
+	require.NoError(t, err)
+
+	// Then: EnvoyFilter should be CREATED for OpenStack
+	shouldDeleteEnvoyFilter := false // OpenStack always creates EnvoyFilter
+	assert.False(t, shouldDeleteEnvoyFilter, "OpenStack should CREATE EnvoyFilter (proxy protocol v1 needed)")
```
kolodziejczak, line 221: I think we assert hardcoded false to be false here

internal/clusterconfig/loadbalancer_test.go
```diff
@@ -0,0 +1,246 @@
...
+
+	isDualStack, err := clusterconfig.IsDualStackEnabled(context.Background(), testClient)
+	require.NoError(t, err)
+
+	// Then: EnvoyFilter should be DELETED (NLB IPv4-only uses DSR, no proxy protocol needed)
+	shouldDeleteEnvoyFilter := shouldUseNLB && !isDualStack
+	assert.True(t, shouldDeleteEnvoyFilter, "NLB IPv4-only should DELETE EnvoyFilter (no proxy protocol needed)")
+}
+
+func TestEnvoyFilterDecision_AWS_NLB_DualStack(t *testing.T) {
```
kolodziejczak, line 166: This test is a bit weird in my opinion, we do not assert a pure output from the tested code. It might be cause of a bad interface or leaky abstraction, but maybe we could revisit it once again.

kolodziejczak (changes requested):

internal/clusterconfig/clustersize_test.go
kubicar: I don't know what then. It is to consider what then (we never had in mind IMHO this kind of scenario)

internal/clusterconfig/loadbalancer_test.go
```diff
@@ -0,0 +1,246 @@
...
+	shouldUseNLB, err := clusterconfig.ShouldUseNLB(context.Background(), testClient)
+	require.NoError(t, err)
+
+	// Simulate dual-stack enabled
+	isDualStack := true
+
+	// Then: EnvoyFilter should be CREATED (NLB dual-stack needs proxy protocol)
+	shouldDeleteEnvoyFilter := shouldUseNLB && !isDualStack
+	assert.False(t, shouldDeleteEnvoyFilter, "NLB dual-stack should CREATE EnvoyFilter (proxy protocol needed)")
+}
```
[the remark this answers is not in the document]
kubicar: Not relevant those tests are prepared for implementation of refactored package.

internal/clusterconfig/loadbalancer_test.go
kubicar: those tests are prepared for implementation of refactored package.

internal/clusterconfig/loadbalancer_test.go
kubicar: those tests are prepared for implementation of refactored package.

internal/clusterconfig/clustersize_test.go
kolodziejczak: yeah, exactly, this is non-blocking/nit, feel free to skip

kolodziejczak (approved):

PR #2144, opened 2026-07-15
docs: adr: restarter component

Ressetkk:
Create proposal for new restarter component.

### DIFF
--- a/docs/contributor/adr/0018-restarter-component.md
+++ b/docs/contributor/adr/0018-restarter-component.md
@@ -1,0 +1,479 @@
+# Restarter component in Istio module
+
+## Status
+
+Proposed
+
+## Context
+
+This document aims to introduce required changes in Istio's restarter component to enhance its functionality and address
+existing limitations.
+
+This document describes the following technical points related to the implementation:
+
+- Lifecycle of the component
+- Features and capabilities of the component
+- Predicate logic for workload restart
+- Event handling and reconciliation process
+- Metrics and observability
+- Available API
+
+This document does NOT enforce the following aspects of the implementation:
+
+- Using specific libraries or frameworks
+- Code placement
+- Design patterns
+
+The libraries and code snippets are suggestions.
+Those final aspects are left to the discretion of the implementer, as long as they do not conflict with the requirements
+and constraints outlined in this document.
+
+### Problem
+
+Current implementation of the restarter component has certain limitations that affect its reliability. These limitations
+include:
+
+- Exponential memory growth due to the number of API calls to the Kubernetes API server, which can lead to performance
+  degradation.
+- Execution during the Istio CR reconciliation, which can cause delays in applying configuration changes and affect the
+  overall responsiveness of the application.
+- Lack of support for customizability for customer to configure and control the restart behavior of workloads, which can
+  lead to unexpected restarts and potential downtime.
+- Code quality issues that can make it difficult to maintain and extend the restarter component, leading to increased
+  technical debt and bugs.
+
+### Background
+
+Restarter component is a critical functionality of istio module that ensures the workloads that are running in the Istio
+service mesh are up to date during updates.
+The first implementation has been introduced in early versions of the module and has been adjusted over time. However,
+with the recent bug reports and investigations, it has been identified that the current implementation is one of the
+major causes of application instability.
+Restarts are handled in one big bang due to the fact that component works within single reconciliation loop of Istio CR.
+Restarting everything in single run can cause massive downtimes which customer does not expect.
+We have also identified that the current implementation causes the Istio module to have exponential memory growth on
+large clusters. The root cause was number of API calls to fetch the list of workloads that are running in the Istio
+service mesh.
+Due to the fact that restarter and Istio CR reconcile loop work as a single binary, the API calls caused the controller
+cache to grow uncontrollably, reaching gigabytes of allocated memory by the controller.
+We had several reports from customers that they have been surprised by the restarts, and it has caused downtime for
+them. There is no way for customer to control the behavior of restarts. This is a major concern and needs to be
+addressed.
+
+## Decision
+
+We have decided to implement a new restarter component that will address the limitations stated above. The new component
+will be designed to be more efficient, reliable, and customizable, while also improving code quality and
+maintainability.
+The new restarter component will be implemented as a separate application that will run independently of the Istio
+reconciliation loop, but the lifecycle of the component will be managed by the Istio module.
+The component will be responsible for monitoring the workloads in the Istio service mesh and restarting them as needed,
+based on the default configuration and overrides provided by the user.
+
+### Lifecycle of the component
+
+To address the memory and performance issues, the new restarter component must be installed as a separate application.
+The component installation and removal will still be managed by the Istio module.
+To handle installation and uninstallation of many Kubernetes resources required by the component (Deployment,
+ServiceAccount, RBAC, Service), we will use Helm. The Istio module will embed a Helm chart and apply it programmatically
+using the Helm Go SDK (`helm.sh/helm/v3`) directly from controller code — no external `helm` binary is required.
+
+This lifecycle is heavily inspired by the Gardener's extensions pattern, which handles the installation of application
+via embedded helm charts.
+
+The Helm chart will be embedded at compile time using Go's `embed.FS`. The installation of a component will be
+integrated into the Istio CR reconciliation loop, and will implement the helm lifecycle:
+
+1. **Install/Upgrade**: When the Istio CR is present and reconciled successfully, the reconciler renders the Helm
+   chart (using `helm/pkg/action.NewInstall` / `action.NewUpgrade`) and applies the resulting manifests via server-side
+   apply.
+2. **Uninstall**: When the Istio CR is deleted (detected via finalizer), the reconciler calls `action.NewUninstall` to
+   tear down the Helm release. This automatically removes all chart-owned resources.
+3. **Drift detection**: The reconciler compares the desired chart values derived from the Istio CR against the
+   last-applied values stored in the Helm release secret. If they differ, an upgrade is triggered.
+
+The Helm release will be stored in the same namespace as the Istio CR (usually `kyma-system`) as a Secret with helm
+labels. This allows standard `helm list` introspection for debugging.
+
+Since Istio version is tightly coupled to the module version, the restarter component version will also be coupled to
+the Istio module version. This means that when the Istio module is updated, the restarter component will also be
+updated, causing subsequent requeue of all workloads by default.
+
+Configuration of the restarter component will be provided via a `.spec.restarter` field in the Istio CR. The reconciler
+will render the Helm chart with values derived from this field, and any changes to it will trigger a Helm upgrade.
+The configuration will be optional, and the component will have a default configuration that is applied when fields are
+not present. The default configuration will be defined in the Helm chart's `values.yaml` file, ensuring that customer
+gets the full functionality without any configuration, but also allowing them to customize the behavior of the component
+if they want to.
+
+```yaml
+spec:
+  restarter:
+    enabled: true
+    logLevel: info
+    reconcileInterval: 15m
+    deployment:
+      autoscaling:
+        enabled: true
+        minReplicas: 1
+        maxReplicas: 3
+      resources:
+        requests:
+          cpu: 100m
+          memory: 128Mi
+        limits:
+          cpu: 200m
+          memory: 256Mi
+```
+
+Changing the lifecycle of the component also requires to adjust the decisions made
+in [ADR-0002](./0002-istio-cr-status-improvements.md) regarding the restart conditions.
+Introducing this change will drop usage of `ProxySidecarRestartSucceeded` condition, and instead introduce a new
+condition `RestarterComponentReady` that will reflect the state of the restarter component. The new condition will be
+set to `True` when the component is installed and running successfully, and `False` when it is not.
+The Istio module will not attempt to restart workloads if the restarter component is not ready, and will log Processing
+message instead.
+If the restarter component failed to start after a certain time, or after a number of restarts, the Istio module will
+set the `RestarterComponentReady` condition to `False`, and will log an Error message.
+From now on, the Istio module will not change the CR status based on the restart status of workloads, but will only
+reflect the state of the restarter component.
+
+| CR state       | Type                             | Status    | Reason                                | Message                                                         |
+|----------------|----------------------------------|-----------|---------------------------------------|-----------------------------------------------------------------|
+| ~~Processing~~ | ~~ProxySidecarRestartSucceeded~~ | ~~True~~  | ~~ProxySidecarRestartSucceeded~~      | ~~Proxy sidecar restart succeeded~~                             |
+| ~~Error~~      | ~~ProxySidecarRestartSucceeded~~ | ~~False~~ | ~~ProxySidecarRestartFailed~~         | ~~Proxy sidecar restart failed~~                                |
+| ~~Processing~~ | ~~ProxySidecarRestartSucceeded~~ | ~~False~~ | ~~ProxySidecarPartiallySucceeded~~    | ~~Proxy sidecar restart partially succeeded~~                   |
+| ~~Warning~~    | ~~ProxySidecarRestartSucceeded~~ | ~~False~~ | ~~ProxySidecarManualRestartRequired~~ | ~~Proxy sidecar manual restart is required for some workloads~~ |
+| Ready          | RestarterComponentReady          | True      | RestarterComponentReady               | Restarter component is installed and running successfully       |
+| Processing     | RestarterComponentReady          | False     | RestarterComponentNotReady            | Restarter component is not ready                                |
+| Error          | RestarterComponentReady          | False     | RestarterComponentFailed              | Restarter component failed to start                             |
+
+### Features and capabilities of the component
+
+We have identified the following features and capabilities that the new restarter component must have.
+
+#### Support for limited workload types
+
+After the analysis of current implementation it's been identified that support for Pods and Jobs was not implemented.
+Instead, the restarter was returning a Warning.
+With ReplicaSet not owned by anything, we delete pod and hope replica controller would heal it.
+This behavior is misleading, as it gives the impression that restarting these workloads is supported, while in reality
+they are not.
+
+To further simplify the decision, the component will support restart only for workloads that are managed by a
+Deployment, StatefulSet or DaemonSet - that means workloads that support rolling restarts.
+To restart workload we will use the similar behavior as `kubectl rollout restart` command, which is to update the pod
+template spec with a new annotation that will trigger a rolling restart of the workload.
+Every time the restart decision is made, the component will update the pod template spec with a new annotation
+`restarter.istio-operator.kyma-project.io/restartedAt`.
+
+If the demand for support of other workload types arises, we can implement it in the future. However, we will not
+support workloads that do not support rolling restarts for now.
+
+#### Maintenance window support
+
+The component must support a maintenance window configuration that allows the user to specify a time range during which
+restarts are allowed.
+This will help to avoid unexpected restarts during critical business hours. The maintenance window will be defined as an
+optional annotation added by the user on the workload.
+For easier configuration, we introduce a new annotation
+`restarter.alpha.istio-operator.kyma-project.io/maintenance-window` that supports a syntax of `Day-of-week HH:MM-HH:MM`
+in UTC.
+
+Parsed as three parts: <day-range> <time-range>. Day range supports:
+
+- Single day: Sat
+- Range: Sat-Sun
+- Comma list: Sat,Sun (if you want non-contiguous days later)
+
+This stays human-readable, requires no cron knowledge, and covers the vast majority of real maintenance window patterns.
+Validation is a single regex:
+
+```go
+// "Sat-Sun 00:00-04:00"
+var windowPattern = regexp.MustCompile(
+`^(Mon|Tue|Wed|Thu|Fri|Sat|Sun)(?:-(Mon|Tue|Wed|Thu|Fri|Sat|Sun))? ([01]\d|2[0-3]):[0-5]\d-([01]\d|2[0-3]):[0-5]\d$`,
+)
+```
+
+The check at reconcile time becomes:
+
+1. Parse day range → is time.Now().Weekday() within it?
+2. Parse time range → is current time-of-day within it?
+3. Both true → proceed with restart; otherwise skip and requeue with `reconcileInterval` until window opens.
+
+#### Restart exclusions
+
+The component must support a configuration that allows the user to specify workloads that should be excluded from
+restarts.
+This will help to avoid restarts of critical workloads that cannot tolerate downtime. We introduce a new annotation
+`restarter.alpha.istio-operator.kyma-project.io/exclude` that supports a boolean value of `true` or `false`. If the
+annotation is set to `true`, the workload will be excluded from restarts.
+If the annotation is not present, the workload will be included in restarts by default.
+
+This mode excludes workload from **all restarts**, including Istio critical CVE proxy updated.
+Using this annotation also implies that the user is aware of the risks of not restarting the workload and take a
+responsibility to keep it updated.
+
+#### Extending the configuration and future-proofing
+
+Those options are defined on workload-level. If there is a need to define them on namespace-level, we can introduce it
+in the future as a separate custom resource and plan migration from workload-level annotations to custom resource with
+label selectors.
+This will allow configuring restarter behavior for multiple workloads with single configuration.
+
+### Predicate logic for workload restart
+
+To decide whether a workload should be restarted, the component uses a set of predicates that are evaluated in order.
+Predicate system is an existing approach and in the long run was a good choice, as it allows to easily add new
+predicates in the future without changing the existing code.
+
+However, current implementation suffers from the unnecessary code complexity and bugs. To address this, a new, static
+approach will be implemented.
+
+Each predicate implements a single `Matcher` interface with a single method `Match()`.
+Workload is a Kubernetes object that implements the `Object` interface. This interface shadows the `client.Object`
+interface from controller-runtime.
+The `Match()` method returns a boolean value indicating whether the workload matches the predicate.
+
+`Match()` returns `true` when the state of the workload is different from the desired state, and the workload should be
+restarted.
+
+```go
+type Object interface {
+client.Object
+}
+
+type Matcher interface {
+Match(obj Object) bool
+}
+```
+
+Predicates only evaluate the workload based on its current state, and should not have any side effects, like fetching
+resource from API.
+Any additional information required for the evaluation should be passed to the predicate via its constructor **before**
+triggering workload evaluation.
+
+The predicates should be extensible in a way that each predicate can be used in different reconciliation loops.
+The list of available predicates must be defined in a single place within a reconcile loop, as a list of `Matcher`
+implementations.
+This list cannot mutate, that means predicates cannot be added or removed at runtime. This is to ensure that the
+predicates are evaluated in a consistent order and that the behavior of the component is predictable.
+
+#### Naming scheme
+
+Predicate names must describe *the condition that makes a restart necessary*, not the action taken — the action is
+always the same (a restart). A predicate returns `true` when the workload has drifted from its desired state, so the
+name
+should read as an assertion of that drift. Two forms are used:
+
+- **`<Subject>Changed`** — for predicates that detect a change relative to what the workload currently reflects. Read as
+  "the subject changed, so the workload is stale". Examples: `CniModeChanged`, `ProxyConfigChanged`,
+  `CompatibilityModeChanged`.
+
+Both forms make the predicate list readable top-to-bottom as a set of restart triggers, and both keep the name focused
+on
+the *why* rather than the *what*. A predicate name should not describe the restart itself (for example, `RestartOnCni`
+or
+`DoProxyRestart` are discouraged) — the restart is implied by any predicate matching.
+
+#### Logical operations of predicates
+
+Each predicate is a self-contained unit that evaluates a single aspect of the workload.
+Predicates can be combined using logical operations to create more complex matching criteria.
+
+By default, evaluation of predicates iterates over a list of defined predicates. This implies the logical OR operation.
+That means if any of the predicates returns `true`, the workload will be restarted.
+
+When a workload must meet 2 or more conditions, `Matcher` implementation can be combined using helper function `And()`.
+
+This function takes a list of `Matcher` implementations and returns a new `Matcher` that evaluates to `true` only if
+all provided `Matcher` implementations evaluate to `true`.
+
+```go
+type andMatcher struct {
+matchers []Matcher
+}
+
+func (a *andMatcher) Match(obj Object) bool {
+for _, matcher := range a.matchers {
+if !matcher.Match(obj) {
+return false
+}
+}
+return true
+}
+
+func And(matchers ...Matcher) Matcher {
+return &andMatcher{matchers: matchers}
+}
+```
+
+### Event handling and reconciliation process
+
+For better stability of a restarter, the implementation would follow the standard Kubernetes controller pattern.
+
+#### Manager and controllers
+
+Each supported workload kind (Deployment, StatefulSet, DaemonSet) has its own dedicated controller-runtime controller.
+All controllers are registered in a single `manager.Manager` instance, sharing the same client, cache, and metrics
+server.
+
+To tackle memory issues, the manager will configure a cache transform on the watched workload types.
+controller-runtime's cache supports a `TransformFunc` that is applied to every object before it is stored in the
+informer cache. By setting a `TransformFunc` on each `ByObject` entry (or globally via `DefaultTransform`), the
+restarter can strip fields that are never used in predicate evaluation — such as `managedFields`, `ownerReferences`, and
+the full pod spec — before the object is committed to memory.
+
+The transform function will retain only the fields required for predicate evaluation. If any of the predicates require
+additional fields, they must be added to the transform function.
+
+**controller-runtime** provides `cache.TransformStripManagedFields()` as a built-in helper that removes `managedFields`
+alone, which is the single largest contributor to object size. For the restarter, a custom `TransformFunc` will be
+registered per object kind via `cache.ByObject.Transform` to retain only the fields listed above, giving tighter memory
+bounds than the built-in helper alone.
+
+#### Kubernetes Events
+
+Kubernetes events API provides easy way to observe the actions performed on the workload component.
+Events provide a way to track any actions made on the customer workload and can serve as a source of messages for
+observability purposes.
+Each controller holds a `record.EventRecorder` obtained at setup via `mgr.GetEventRecorderFor("istio-restarter")`. All
+emitted Events share the same `reportingComponent` field, making them easy to filter:
+
+```
+kubectl get events --field-selector reportingComponent=istio-restarter
+```
+
+Events are emitted using `recorder.Event(obj, eventType, reason, message)` or `recorder.Eventf(...)` for formatted
+messages. controller-runtime deduplicates events with the same reason and message within a short window, so a
+high-frequency reconcile loop does not flood the Events API.
+
+`RestartFailed` event is emitted when the restart operation itself fails, for example due to an API error while patching
+the workload. Configuration problems such as a malformed maintenance window are reported through the dedicated
+`RestartConfigInvalid` event and the `invalid_window` skip metric, not as a `RestartFailed` event.
+
+`RestartRequired` event is emitted when a predicate would have matched but is suppressed by policy (for example, an
+excluded workload that is running a stale proxy).
+
+`RestartConfigInvalid` event is emitted by the preflight check when a workload's restarter configuration annotation is
+malformed (for example, a maintenance window that does not match the expected syntax). It carries the workload identity
+and the parse error so the user can locate and fix the annotation. It is deduplicated by Kubernetes, so a persistently
+invalid annotation does not spam the Events API.
+
+| Reason                 | Type    | Trigger                                          |
+|------------------------|---------|--------------------------------------------------|
+| `RestartSuccess`       | Normal  | Workload successfully restarted                  |
+| `RestartFailed`        | Warning | Workload failed to restart                       |
+| `RestartRequired`      | Warning | Workload requires manual restart                 |
+| `RestartConfigInvalid` | Warning | Workload has a malformed restarter configuration |
+
+#### Reconciliation decision diagram
+
+```
+Workload event (create / update / requeue)
+           │
+           ▼
+    Fetch workload
+    (NotFound → ignore)
+           │
+           ▼
+  ┌──────────────────┐
+  │ preflight check  │──deny──► skip + requeue, bump skipped_total{reason}
+  │ (maint. window)  │          (RestartConfigInvalid event if malformed)
+  └────────┬─────────┘
+           │ allow
+           ▼
+  ┌──────────────────┐
+  │ any Matcher      │──no──► no-op, no event
+  │ returns true?    │
+  └────────┬─────────┘
+           │ yes (restart warranted)
+           ▼
+  ┌──────────────────┐         set restart_pending, bump skipped_total{reason=excluded},
+  │ excluded?        │──yes──► emit RestartRequired event, do NOT restart
+  └────────┬─────────┘
+           │ no
+           ▼
+      do Restart()
+           │
+      ┌────┴────┐
+      │ error?  │
+      └────┬────┘
+     yes   │    no
+      ▼    │    ▼
+  Warning  │  Normal Event
+  Event    │  "Restarted"
+  requeue  │  return OK
+```
+
+### Metrics and observability
+
+The restarter component exposes a Prometheus metrics endpoint (`/metrics`) served by the controller-runtime manager.
+All custom metrics are prefixed `istio_restarter_`. In addition to the metrics below, the manager automatically exports
+the standard controller-runtime metrics (`controller_runtime_reconcile_total`,
+`controller_runtime_reconcile_errors_total`,
+`controller_runtime_reconcile_time_seconds`, `workqueue_depth`, `workqueue_adds_total`, `workqueue_retries_total`),
+which
+already cover reconcile throughput, error rate, and queue depth per controller — so the custom metrics below focus on
+restarter-specific decisions rather than duplicating those.
+
+**Cardinality note:** labels are restricted to bounded-cardinality dimensions (`kind`, `reason`, `result`). Per-workload
+identity (namespace/name) is intentionally *not* a label — on a large cluster that would produce unbounded time series
+and reintroduce the memory problem this ADR set out to solve. Per-workload restart facts are surfaced through Kubernetes
+Events instead (filterable by `reportingComponent=istio-restarter`).
+
+| Metric                                                  | Type      | Labels                                                        | Description                                                                                                                                                                                                                                                               |
+|---------------------------------------------------------|-----------|---------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `istio_restarter_restarts_total`                        | Counter   | `kind`, `result` (`success` / `error`)                        | Restart attempts, split by outcome. Restart throughput and failure rate derive from this single metric.                                                                                                                                                                   |
+| `istio_restarter_skipped_total`                         | Counter   | `reason` (`excluded`, `maintenance_window`, `invalid_window`) | Workloads skipped by the preflight check or exclusion. Mirrors the skip reasons one-to-one.                                                                                                                                                                               |
+| `istio_restarter_restart_pending`                       | Gauge     | `kind`                                                        | Workloads a predicate matched but the exclusion check suppressed (a restart is *required* but not performed). Should trend to zero after a maintenance window opens; a stuck non-zero value means restarts are being blocked. Corresponds to the `RestartRequired` event. |
+| `istio_restarter_predicate_evaluation_duration_seconds` | Histogram | `kind`                                                        | Time to evaluate the preflight check and predicate chain for one workload. Detects predicates that became expensive.                                                                                                                                                      |
+| `istio_restarter_maintenance_window_active`             | Gauge     | —                                                             | `1` while *any* configured maintenance window is currently open, else `0`. A cluster-wide debugging signal for "why did nothing restart" — deliberately unlabeled to stay bounded-cardinality.                                                                            |
+| `istio_restarter_component_info`                        | Gauge     | `version`, `istio_version`                                    | Constant `1`; carries the restarter build version and the target Istio version as labels. Lets dashboards correlate restart activity with a specific Istio upgrade.                                                                                                       |
+
+## Consequences
+
+Accepting this decision resolves the core problems identified in the current implementation and introduces a set of
+improvements that make the restarter component more reliable, maintainable, and customer-friendly.
+
+- **Memory and performance**: The current implementation executes entirely within the Istio CR reconciliation loop.
+  Every reconcile cycle issues direct API calls to list all Pods across the cluster, bypassing the controller cache and
+  causing the cache to grow uncontrollably — reaching gigabytes on large clusters. The new component runs as a separate
+  process with its own informer cache and a `TransformFunc` that strips unused fields before objects are stored. This
+  bounds memory usage to the minimum fields required for predicate evaluation and eliminates the impact on the main
+  controller's memory footprint entirely.
+- **Restart behavior**: The current implementation restarts all matching workloads in a single reconcile pass — a "big
+  bang" approach that has caused customer-visible downtime during Istio upgrades. The new component replaces this with a
+  controller-runtime reconcile loop per workload kind, where each workload is processed individually as an event
+  arrives. Restarts are performed via annotation patches, which delegate the actual
+  rollout to the Deployment/StatefulSet/DaemonSet controller and respect PodDisruptionBudgets. This eliminates the
+  mass-eviction risk entirely.
+- **Customer control**: The current implementation provides no mechanism for customers to influence when or whether
+  their workloads are restarted. The new component introduces a per-workload exclude annotation and a maintenance window
+  annotation, giving customers explicit control over both opt-out and scheduling of restarts. These are intentionally
+  designed to require no Istio or cron expertise — a plain `Sat-Sun 00:00-04:00` string is sufficient.
+- **Supported workload types**: The current restarter nominally handles Pods and Jobs but returns only a Warning for
+  them, silently skipping restarts. Bare ReplicaSets are handled by deleting the pod and relying on the replication
+  controller to heal, which is fragile and non-obvious. The new component explicitly restricts support to Deployment,
+  StatefulSet, and DaemonSet — the only kinds that support rolling restarts — and rejects the ambiguity by design. This
+  simplifies the predicate evaluation surface and removes a class of silent failures.
+- **Predicate system**: The existing predicate system carries two interface types (`SidecarProxyPredicate` for pod-level
+  evaluation and `IngressGatewayPredicate` for gateway evaluation) with inconsistent contracts and known bugs. The new
+  implementation replaces both with a single `Matcher` interface that operates on a `client.Object`, is stateless, and
+  receives all required context via its constructor. Predicates can be shared across all workload-kind controllers
+  without modification, and the immutable matcher list prevents runtime state mutations that were a source of bugs in
+  the prior implementation.
+- **Observability**: The current implementation surfaces restart outcomes only through Istio CR status conditions (
+  `ProxySidecarRestartSucceeded` and its variants), which provide a single aggregate view and are reset on every
+  reconcile. The new component replaces these with a `RestarterComponentReady` condition that reflects the health of the
+  component itself, and complements it with per-workload Kubernetes Events (filterable by
+  `reportingComponent=istio-restarter`) and a set of Prometheus metrics covering throughput, errors, queue depth, and
+  maintenance window state. This gives both operators and customers multiple independent signals to diagnose restart
+  behavior without accessing controller logs.
+- **Operational complexity**: The separation into a standalone application introduces new concerns: a second binary and
+  Helm chart to build, publish, and version; Helm release state that must be cleaned up on uninstall via the Istio CR
+  finalizer; and a separate informer cache that watches some of the same resource types as the main controller. These
+  are real costs, but they are bounded and predictable. The Helm-based lifecycle provides a standard inventory of all
+  owned resources, making installation and removal auditable. The trade-off is judged acceptable given that the memory
+  and reliability problems it solves have directly caused production incidents.
--- a/docs/contributor/adr/0018-restarter-component.md
+++ b/docs/contributor/adr/0018-restarter-component.md
@@ -53,7 +53,14 @@ Restarts are handled in one big bang due to the fact that component works within
 Restarting everything in single run can cause massive downtimes which customer does not expect.
 We have also identified that the current implementation causes the Istio module to have exponential memory growth on
 large clusters. The root cause was number of API calls to fetch the list of workloads that are running in the Istio
-service mesh.
+service mesh. To admit a workload to restart, be it a Deployment, StatefulSet or DaemonSet, restarter works on the list
+of uncorrelated pods.
+In single reconciliation loop, restarter fetches the list of all pods in a cluster in paginated chunks of 100. In each
+chunk restarter admits the pods to the predicates, which usually requires additional API calls to retrieve the parent of
+admitted pod during restart. Listing all pods in a cluster in a single reconcile loop, even with pagination, costs a lot
+of memory. And with larger clusters we might end up with stale cache which can cause unwanted API errors and another run
+of restart.
+
 Due to the fact that restarter and Istio CR reconcile loop work as a single binary, the API calls caused the controller
 cache to grow uncontrollably, reaching gigabytes of allocated memory by the controller.
 We had several reports from customers that they have been surprised by the restarts, and it has caused downtime for
@@ -65,10 +72,16 @@ addressed.
 We have decided to implement a new restarter component that will address the limitations stated above. The new component
 will be designed to be more efficient, reliable, and customizable, while also improving code quality and
 maintainability.
-The new restarter component will be implemented as a separate application that will run independently of the Istio
-reconciliation loop, but the lifecycle of the component will be managed by the Istio module.
+The new restarter component will be implemented as a separate deployment that will run independently of the Istio
+reconciliation loop, but the lifecycle of it will be managed by the Istio module.
 The component will be responsible for monitoring the workloads in the Istio service mesh and restarting them as needed,
 based on the default configuration and overrides provided by the user.
+
+Because the new restarter will work on the asynchronous event-driven model, it will fetch the list of Pods from
+"top-to-bottom". This updated behavior starts from the parent workload (Deployment, StatefulSet, DaemonSet) and then
+uses selectors to retrieve all child Pods to admit.
+This will greatly reduce the number of API calls and sizes of the objects to admit in single loop. It will also ensure
+that admitted workload is up-to-date with the current cache state.
 
 ### Lifecycle of the component
 
@@ -128,7 +141,7 @@ spec:
 
 Changing the lifecycle of the component also requires to adjust the decisions made
 in [ADR-0002](./0002-istio-cr-status-improvements.md) regarding the restart conditions.
-Introducing this change will drop usage of `ProxySidecarRestartSucceeded` condition, and instead introduce a new
+Introducing this change will **drop** usage of `ProxySidecarRestartSucceeded` condition, and instead introduce a new
 condition `RestarterComponentReady` that will reflect the state of the restarter component. The new condition will be
 set to `True` when the component is installed and running successfully, and `False` when it is not.
 The Istio module will not attempt to restart workloads if the restarter component is not ready, and will log Processing
@@ -177,25 +190,18 @@ restarts are allowed.
 This will help to avoid unexpected restarts during critical business hours. The maintenance window will be defined as an
 optional annotation added by the user on the workload.
 For easier configuration, we introduce a new annotation
-`restarter.alpha.istio-operator.kyma-project.io/maintenance-window` that supports a syntax of `Day-of-week HH:MM-HH:MM`
+`restarter.istio-operator.kyma-project.io/maintenance-window` that supports a syntax of `Day-of-week HH:MM-HH:MM`
 in UTC.
 
-Parsed as three parts: <day-range> <time-range>. Day range supports:
+Parsed as two parts: <day-range> <time-range>. Day range supports:
 
 - Single day: Sat
 - Range: Sat-Sun
 - Comma list: Sat,Sun (if you want non-contiguous days later)
 
 This stays human-readable, requires no cron knowledge, and covers the vast majority of real maintenance window patterns.
-Validation is a single regex:
 
-```go
-// "Sat-Sun 00:00-04:00"
-var windowPattern = regexp.MustCompile(
-`^(Mon|Tue|Wed|Thu|Fri|Sat|Sun)(?:-(Mon|Tue|Wed|Thu|Fri|Sat|Sun))? ([01]\d|2[0-3]):[0-5]\d-([01]\d|2[0-3]):[0-5]\d$`,
-)
-```
-
+Validation parses annotation and rejects invalid syntax by emitting appropriate Event.
 The check at reconcile time becomes:
 
 1. Parse day range → is time.Now().Weekday() within it?
@@ -207,11 +213,11 @@ The check at reconcile time becomes:
 The component must support a configuration that allows the user to specify workloads that should be excluded from
 restarts.
 This will help to avoid restarts of critical workloads that cannot tolerate downtime. We introduce a new annotation
-`restarter.alpha.istio-operator.kyma-project.io/exclude` that supports a boolean value of `true` or `false`. If the
+`restarter.istio-operator.kyma-project.io/exclude` that supports a boolean value of `true` or `false`. If the
 annotation is set to `true`, the workload will be excluded from restarts.
 If the annotation is not present, the workload will be included in restarts by default.
 
-This mode excludes workload from **all restarts**, including Istio critical CVE proxy updated.
+This mode excludes workload from **all restarts**, including Istio updates that mitigate a CVE in the proxy.
 Using this annotation also implies that the user is aware of the risks of not restarting the workload and take a
 responsibility to keep it updated.
 
@@ -271,11 +277,9 @@ should read as an assertion of that drift. Two forms are used:
   "the subject changed, so the workload is stale". Examples: `CniModeChanged`, `ProxyConfigChanged`,
   `CompatibilityModeChanged`.
 
-Both forms make the predicate list readable top-to-bottom as a set of restart triggers, and both keep the name focused
-on
-the *why* rather than the *what*. A predicate name should not describe the restart itself (for example, `RestartOnCni`
-or
-`DoProxyRestart` are discouraged) — the restart is implied by any predicate matching.
+The form makes the predicate list readable top-to-bottom as a set of restart triggers, and both keep the name focused
+on *why* rather than *what*. A predicate name should not describe the restart itself (for example, `RestartOnCni`
+or `DoProxyRestart` are discouraged) — the restart is implied by any predicate matching.
 
 #### Logical operations of predicates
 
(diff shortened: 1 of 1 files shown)


docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,479 @@
...
+The component will be responsible for monitoring the workloads in the Istio service mesh and restarting them as needed,
+based on the default configuration and overrides provided by the user.
+
+### Lifecycle of the component
+
+To address the memory and performance issues, the new restarter component must be installed as a separate application.
+The component installation and removal will still be managed by the Istio module.
+To handle installation and uninstallation of many Kubernetes resources required by the component (Deployment,
+ServiceAccount, RBAC, Service), we will use Helm. The Istio module will embed a Helm chart and apply it programmatically
+using the Helm Go SDK (`helm.sh/helm/v3`) directly from controller code — no external `helm` binary is required.
```
barchw: Although using Helm is potentially an improvement, I would not consider it a requirement for the sidecar restarter part. That is: it solves problems that are already solved in the existing logic.

The lifecycle/Helm section is a deployment decision independent of the restart logic. You could adopt the restart redesign and use it with the repo's existing mechanism without changing the installation logic.

I would recommend to extract that into a seperate ADR if we want to consider that, or drop that entirely. This part has very significant overall impact and implementation cost.

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,479 @@
...
+1. Parse day range → is time.Now().Weekday() within it?
+2. Parse time range → is current time-of-day within it?
+3. Both true → proceed with restart; otherwise skip and requeue with `reconcileInterval` until window opens.
+
+#### Restart exclusions
+
+The component must support a configuration that allows the user to specify workloads that should be excluded from
+restarts.
+This will help to avoid restarts of critical workloads that cannot tolerate downtime. We introduce a new annotation
+`restarter.alpha.istio-operator.kyma-project.io/exclude` that supports a boolean value of `true` or `false`. If the
```
barchw: I think this should rather be a label instead of annotation.
This would make it easy to exclude in matching logic and directly follows Istio's injection label approach.

docs/contributor/adr/0018-restarter-component.md
barchw, line 1: The decision is missing the explanation to the idea behind going from the top (i.e. from Deployment, DaemonSet and StatefullSet) instead of from the bottom (the actual Pod).
This is a good improvement, but needs to explain how the actual evaluation will happen -> going to the Pod level is still unavoidable, as there is no information about the version, config etc. of the Istio proxy directly on the parent.

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,479 @@
...
+
+To further simplify the decision, the component will support restart only for workloads that are managed by a
+Deployment, StatefulSet or DaemonSet - that means workloads that support rolling restarts.
+To restart workload we will use the similar behavior as `kubectl rollout restart` command, which is to update the pod
+template spec with a new annotation that will trigger a rolling restart of the workload.
+Every time the restart decision is made, the component will update the pod template spec with a new annotation
+`restarter.istio-operator.kyma-project.io/restartedAt`.
+
+If the demand for support of other workload types arises, we can implement it in the future. However, we will not
+support workloads that do not support rolling restarts for now.
```
barchw: There is no guarantee that "the demand" is not already there.
Dropping support here, even if the current behavior is unjustified contributes a breaking change to the design.

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,479 @@
...
+}
+
+type Matcher interface {
+Match(obj Object) bool
+}
+```
+
+Predicates only evaluate the workload based on its current state, and should not have any side effects, like fetching
+resource from API.
+Any additional information required for the evaluation should be passed to the predicate via its constructor **before**
```
barchw: Does the additional information mean we pass in the Pods owned by the Object?

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,479 @@
...
+Predicate names must describe *the condition that makes a restart necessary*, not the action taken — the action is
+always the same (a restart). A predicate returns `true` when the workload has drifted from its desired state, so the
+name
+should read as an assertion of that drift. Two forms are used:
+
+- **`<Subject>Changed`** — for predicates that detect a change relative to what the workload currently reflects. Read as
+  "the subject changed, so the workload is stale". Examples: `CniModeChanged`, `ProxyConfigChanged`,
+  `CompatibilityModeChanged`.
+
+Both forms make the predicate list readable top-to-bottom as a set of restart triggers, and both keep the name focused
```
barchw:
```suggestion
The form makes the predicate list readable top-to-bottom as a set of restart triggers, and both keep the name focused
```

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,479 @@
...
+spec:
+  restarter:
+    enabled: true
+    logLevel: info
+    reconcileInterval: 15m
+    deployment:
+      autoscaling:
+        enabled: true
+        minReplicas: 1
+        maxReplicas: 3
```
barchw: If autoscaling is an option, the impact should be considered.
This can potentially result in race conditions, or it might not provide any value because of LeaderElection happening.

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,479 @@
...
+
+The component must support a maintenance window configuration that allows the user to specify a time range during which
+restarts are allowed.
+This will help to avoid unexpected restarts during critical business hours. The maintenance window will be defined as an
+optional annotation added by the user on the workload.
+For easier configuration, we introduce a new annotation
+`restarter.alpha.istio-operator.kyma-project.io/maintenance-window` that supports a syntax of `Day-of-week HH:MM-HH:MM`
+in UTC.
+
+Parsed as three parts: <day-range> <time-range>. Day range supports:
```
barchw:
```suggestion
Parsed as two parts: <day-range> <time-range>. Day range supports:
```

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,479 @@
...
+This stays human-readable, requires no cron knowledge, and covers the vast majority of real maintenance window patterns.
+Validation is a single regex:
+
+```go
+// "Sat-Sun 00:00-04:00"
+var windowPattern = regexp.MustCompile(
+`^(Mon|Tue|Wed|Thu|Fri|Sat|Sun)(?:-(Mon|Tue|Wed|Thu|Fri|Sat|Sun))? ([01]\d|2[0-3]):[0-5]\d-([01]\d|2[0-3]):[0-5]\d$`,
+)
+```
+
```
barchw: This validation does not check whether the first range is a range happening before the latter.

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,479 @@
...
+support workloads that do not support rolling restarts for now.
+
+#### Maintenance window support
+
+The component must support a maintenance window configuration that allows the user to specify a time range during which
+restarts are allowed.
+This will help to avoid unexpected restarts during critical business hours. The maintenance window will be defined as an
+optional annotation added by the user on the workload.
+For easier configuration, we introduce a new annotation
+`restarter.alpha.istio-operator.kyma-project.io/maintenance-window` that supports a syntax of `Day-of-week HH:MM-HH:MM`
```
barchw: This might be handled as a label.

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,479 @@
...
+var windowPattern = regexp.MustCompile(
+`^(Mon|Tue|Wed|Thu|Fri|Sat|Sun)(?:-(Mon|Tue|Wed|Thu|Fri|Sat|Sun))? ([01]\d|2[0-3]):[0-5]\d-([01]\d|2[0-3]):[0-5]\d$`,
+)
+```
+
+The check at reconcile time becomes:
+
+1. Parse day range → is time.Now().Weekday() within it?
+2. Parse time range → is current time-of-day within it?
+3. Both true → proceed with restart; otherwise skip and requeue with `reconcileInterval` until window opens.
```
barchw: This does not coexist well with the fact that the maintenance window is defined on workload rather than globally and has potential to skip over workloads entirely.
The reconcile interval has to then be sufficiently small (this is not guaranteed as it is configurable in the design), and the maintenance window has to be sufficiently large for the reconciliation to work.

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,479 @@
...
+
+If the demand for support of other workload types arises, we can implement it in the future. However, we will not
+support workloads that do not support rolling restarts for now.
+
+#### Maintenance window support
+
+The component must support a maintenance window configuration that allows the user to specify a time range during which
+restarts are allowed.
+This will help to avoid unexpected restarts during critical business hours. The maintenance window will be defined as an
+optional annotation added by the user on the workload.
```
barchw: Adding it on workload implies that the calculation happens at the time the workload is considered for a restart.
In a worse case scenario, on a globally triggered rollout, this might mean that some subset of workloads that have the same maintenance window configured are / aren't restarted if the reconciliation run exceeds the window boundaries (i.e. starts before left bound or ends after right bound)
I don't know whether this would be the desired outcome.

docs/contributor/adr/0018-restarter-component.md
barchw: Why use "alpha"?

docs/contributor/adr/0018-restarter-component.md
barchw: Why use alpha?

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,479 @@
...
+#### Restart exclusions
+
+The component must support a configuration that allows the user to specify workloads that should be excluded from
+restarts.
+This will help to avoid restarts of critical workloads that cannot tolerate downtime. We introduce a new annotation
+`restarter.alpha.istio-operator.kyma-project.io/exclude` that supports a boolean value of `true` or `false`. If the
+annotation is set to `true`, the workload will be excluded from restarts.
+If the annotation is not present, the workload will be included in restarts by default.
+
+This mode excludes workload from **all restarts**, including Istio critical CVE proxy updated.
```
barchw:
```suggestion
This mode excludes workload from **all restarts**, including Istio updates that mitigate a CVE in the proxy.
```

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,479 @@
...
+|------------------------|---------|--------------------------------------------------|
+| `RestartSuccess`       | Normal  | Workload successfully restarted                  |
+| `RestartFailed`        | Warning | Workload failed to restart                       |
+| `RestartRequired`      | Warning | Workload requires manual restart                 |
+| `RestartConfigInvalid` | Warning | Workload has a malformed restarter configuration |
+
+#### Reconciliation decision diagram
+
+```
+Workload event (create / update / requeue)
```
barchw: I think the idea of triggering on workload changes is not fully explained in here (in the overall ADR). This per workload reconcile approach does not work well with the intent of triggering on helm changes, as this change happens globally. 

What needs to be explained is how and when does the reconciliation actually trigger. Does a global (Istio CR) change trigger reconcile for all workloads, or does it not do that with the intent of waiting for a time based requeue? If it is triggered, how does it know what actually needs to be evaluated as config change history is lost after apply (even in helm case)?

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,479 @@
...
+           │
+           ▼
+    Fetch workload
+    (NotFound → ignore)
+           │
+           ▼
+  ┌──────────────────┐
+  │ preflight check  │──deny──► skip + requeue, bump skipped_total{reason}
+  │ (maint. window)  │          (RestartConfigInvalid event if malformed)
```
barchw: This depends heavily on how will the reconciliation be triggered in case of a globally triggered restart.

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,479 @@
...
+The Istio module will not attempt to restart workloads if the restarter component is not ready, and will log Processing
+message instead.
+If the restarter component failed to start after a certain time, or after a number of restarts, the Istio module will
+set the `RestarterComponentReady` condition to `False`, and will log an Error message.
+From now on, the Istio module will not change the CR status based on the restart status of workloads, but will only
+reflect the state of the restarter component.
+
+| CR state       | Type                             | Status    | Reason                                | Message                                                         |
+|----------------|----------------------------------|-----------|---------------------------------------|-----------------------------------------------------------------|
+| ~~Processing~~ | ~~ProxySidecarRestartSucceeded~~ | ~~True~~  | ~~ProxySidecarRestartSucceeded~~      | ~~Proxy sidecar restart succeeded~~                             |
```
barchw: What component is the owner of `ProxySidecarRestartSucceeded` Condition?
When will it be set with Ready: True?

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,479 @@
...
+We had several reports from customers that they have been surprised by the restarts, and it has caused downtime for
+them. There is no way for customer to control the behavior of restarts. This is a major concern and needs to be
+addressed.
+
+## Decision
+
+We have decided to implement a new restarter component that will address the limitations stated above. The new component
+will be designed to be more efficient, reliable, and customizable, while also improving code quality and
+maintainability.
+The new restarter component will be implemented as a separate application that will run independently of the Istio
```
barchw: What does "application" mean?
Is it a:
- container
- deployment
- process in the same container?

barchw (changes requested):
The thing that needs the most clarification is https://github.com/kyma-project/istio/pull/2144/changes#r3616522217

docs/contributor/adr/0018-restarter-component.md
Ressetkk: I think the way how it would be installed on a cluster, should be defined here as a complete solution, because we didn't agree on anything global-scope yet. Next ADR can extend existing decision record IMO.

docs/contributor/adr/0018-restarter-component.md
Ressetkk: Labels are used for selectors. Generally for this kind of implementation annotations are better suited, as they do not possibly affect selectors, only metadata.

docs/contributor/adr/0018-restarter-component.md
Ressetkk: The implementation for Pods/Jobs have not been updated for years, and only return "Warning" when you encounter them. It really don't think we should support such scenario for now.

I can rephrase it that it would be a breaking change to the current design, but I still insist on not implementing non-rollout supported workloads, as it causes downtime.

docs/contributor/adr/0018-restarter-component.md
Ressetkk: Yes. Fetch the pods before, pass them as objects to the predicate.

docs/contributor/adr/0018-restarter-component.md
Ressetkk: Most likely will not provide any benefits, and the VPA would be more suitable here. Having replicas for a controller is rather keeping a hot spare in case of crash.

docs/contributor/adr/0018-restarter-component.md
Ressetkk: Labels are used for selectors. Generally for this kind of implementation annotations are better suited, as they do not possibly affect selectors, only metadata.

docs/contributor/adr/0018-restarter-component.md
Ressetkk: The idea was simple - maintenance windows are for the customer to steer when the workload is supposed to restart. With the static `reconcileInterval`, eg. `15m` we should check every 15 min if the workload can be updated. 

We can have a global config introduced as a singleton, and do a simple comparison: if the maintenance window is smaller than `reconcileInterval`. If it is, then return an error and project `RestarterConfigInvalid` event.

docs/contributor/adr/0018-restarter-component.md
Ressetkk: Because it's a new feature, that might evolve in the long run.

docs/contributor/adr/0018-restarter-component.md
Ressetkk: Because it's a new feature, that might evolve in the long run.

docs/contributor/adr/0018-restarter-component.md
Ressetkk: By "application" I mean the separate binary living alongside an Istio-operator independently from its pod. Most likely we will use deployment.

docs/contributor/adr/0018-restarter-component.md
Ressetkk: The idea is to **drop** this predicate as it will be invalid - istio operator will NOT do any restarts.

docs/contributor/adr/0018-restarter-component.md
barchw: This is not enough of a justification as a migration will be required from users because of this choice

docs/contributor/adr/0018-restarter-component.md
Ressetkk: Globally triggered restarts happen when the controller gets redeployed. This is the main idea to not have globally triggered restarts, and have workloads gradually be restarted over the `reconcileInterval` window.

docs/contributor/adr/0018-restarter-component.md
Ressetkk: It really depends. The overall idea is to not have global restarts that would cause a massive amount of workloads to be restarted at once, rather have them gradually and asynchronously rerolled. in `reconcileInterval` interval. Predicates should define whether the state of the workload differs either by Istio CR changed, or the actual workload has some different configurations on the single workload event.

The other thing is changes that affect restarter redeployment. Since restarter and module are bundled, every upgrade of a module will cause restarter to be redeployed too, causing massive restart as a side effect, due to all resources being reconciled on startup. I think it's better to introduce that this side effect occurs. Additionally, any changes to the restarter config will cause it to be redeployed either way.

docs/contributor/adr/0018-restarter-component.md
barchw: No, this is out-of-scope for restarter.
If we want to agree on install, that's where a second ADR should be created. This will keep the discussion of this installation solution out of this PR.

docs/contributor/adr/0018-restarter-component.md
Ressetkk: The config changes are still parts that I haven't figured out completely. I was thinking about re-using last-applied-config approach, or some config hashing which we had in other module.

docs/contributor/adr/0018-restarter-component.md
barchw: I agree, I think the warning can be removed to be honest.
The owned by replicaset deletion should stay though in my opinion.

docs/contributor/adr/0018-restarter-component.md
Ressetkk: I completely understand your reasoning, but for the scope of the entire component, this discussion is needed. I don't think we have time to introduce yet another ADR for the sake of "generic component lifecycle" for now.

docs/contributor/adr/0018-restarter-component.md
barchw: Why should the exclusion not be treated as a selector?

docs/contributor/adr/0018-restarter-component.md
barchw: I think this is OK to use helm if only scoped to the install/apply of the restarter component. This was not clear to me initially.

barchw: I guess it should be also defined what actual Kubernetes Objects potentially will be used, as the config already uses the naming `deployment`.

docs/contributor/adr/0018-restarter-component.md
Ressetkk: I would be very cautious about it. ReplicaSets are usually used as subresources for other resources (eg. Deployments). That means this configuration would apply for ReplicaSets of those resources. Also, blindly removing a pod and hope for ReplicaSet to recreate it can cause downtime. 

Consider one ReplicaSet with single Pod and without a parent. You kill one pod and cause a downtime which basically results in critical data loss. It's really a miracle we didn't get any major complaints about that, or no one is really using restarts with bare ReplicaSets.

docs/contributor/adr/0018-restarter-component.md
Ressetkk: I left it that way on purpose, because with helm this configuration should be based on the `values.yaml` of the corresponding helm chart. It's the same way how `ProviderConfig` in gardener extension is working.

docs/contributor/adr/0018-restarter-component.md
Ressetkk: You don't want to have the custom metadata provided by a client to be a part of a selector. Users can accidentally modify, override or even delete the labels which can cause downtime on their side, as label selectors are generally used by Services, VirtualServices and other implementations. Since annotations are purely for metadata, they are better suited for this kind of configuration.

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,483 @@
...
+The component must support a configuration that allows the user to specify workloads that should be excluded from
+restarts.
+This will help to avoid restarts of critical workloads that cannot tolerate downtime. We introduce a new annotation
+`restarter.istio-operator.kyma-project.io/exclude` that supports a boolean value of `true` or `false`. If the
+annotation is set to `true`, the workload will be excluded from restarts.
+If the annotation is not present, the workload will be included in restarts by default.
+
+This mode excludes workload from **all restarts**, including Istio updates that mitigate a CVE in the proxy.
+Using this annotation also implies that the user is aware of the risks of not restarting the workload and take a
+responsibility to keep it updated.
```
kubicar: IMO maintenance window should be enough, this exclusion potentially can bring more troubles than benefits. We should count, that in kubernetes we should not promote this kind of workloads (for example replace during node upgrade will happen why for Istio restarts scenario should be different? Let's don't introduce something which will be useless)

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,483 @@
...
+The first implementation has been introduced in early versions of the module and has been adjusted over time. However,
+with the recent bug reports and investigations, it has been identified that the current implementation is one of the
+major causes of application instability.
+Restarts are handled in one big bang due to the fact that component works within single reconciliation loop of Istio CR.
+Restarting everything in single run can cause massive downtimes which customer does not expect.
+We have also identified that the current implementation causes the Istio module to have exponential memory growth on
+large clusters. The root cause was number of API calls to fetch the list of workloads that are running in the Istio
+service mesh. To admit a workload to restart, be it a Deployment, StatefulSet or DaemonSet, restarter works on the list
+of uncorrelated pods.
+In single reconciliation loop, restarter fetches the list of all pods in a cluster in paginated chunks of 100. In each
```
mluk-sap: We may also mention that the pagination was introduced in order to not cause high load on API Server.

And the main issue is that the cursor-based pagination is valid only for a short time (~5 minutes), so the current implementation is very fragile (cursor validity vs data plane restart time)

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,483 @@
...
+   tear down the Helm release. This automatically removes all chart-owned resources.
+3. **Drift detection**: The reconciler compares the desired chart values derived from the Istio CR against the
+   last-applied values stored in the Helm release secret. If they differ, an upgrade is triggered.
+
+The Helm release will be stored in the same namespace as the Istio CR (usually `kyma-system`) as a Secret with helm
+labels. This allows standard `helm list` introspection for debugging.
+
+Since Istio version is tightly coupled to the module version, the restarter component version will also be coupled to
+the Istio module version. This means that when the Istio module is updated, the restarter component will also be
+updated, causing subsequent requeue of all workloads by default.
```
mluk-sap: > causing subsequent requeue of all workloads by default
I think it goes to far here. The paragraph describes how the restarter component is versioned. Requeue of all workloads - is a consequence of its 'business logic' and it is unrelated to the versioning model.

docs/contributor/adr/0018-restarter-component.md
mluk-sap: The restarter component is not a critical functionality, so it doesn't need any 'hot spares'.
AFAIK, we were even thinking about stopping it during Istio upgrade to avoid race conditions.

docs/contributor/adr/0018-restarter-component.md
mluk-sap: Maybe we should also make it clear that it is only for 'data plane' restart, which does NOT cover Istio controller and its directly managed components, like Ingress.

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,483 @@
...
+### Features and capabilities of the component
+
+We have identified the following features and capabilities that the new restarter component must have.
+
+#### Support for limited workload types
+
+After the analysis of current implementation it's been identified that support for Pods and Jobs was not implemented.
+Instead, the restarter was returning a Warning.
+With ReplicaSet not owned by anything, we delete pod and hope replica controller would heal it.
+This behavior is misleading, as it gives the impression that restarting these workloads is supported, while in reality
```
mluk-sap: I'm loudly thinking about splitting it into two **separate** mechanisms:
- one mechanism doing the upgrade
- separate one that analyzes the istio proxy versions

Because in case of bare pods / pod owned by jobs we currently fulfill the second - we just report that we see an outdated version.

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,483 @@
...
+3. Both true → proceed with restart; otherwise skip and requeue with `reconcileInterval` until window opens.
+
+#### Restart exclusions
+
+The component must support a configuration that allows the user to specify workloads that should be excluded from
+restarts.
+This will help to avoid restarts of critical workloads that cannot tolerate downtime. We introduce a new annotation
+`restarter.istio-operator.kyma-project.io/exclude` that supports a boolean value of `true` or `false`. If the
+annotation is set to `true`, the workload will be excluded from restarts.
+If the annotation is not present, the workload will be included in restarts by default.
```
kubicar: then, how we will guarantee that data plane is in sync / how customer will know that workload need restart.

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,483 @@
...
+
+### Features and capabilities of the component
+
+We have identified the following features and capabilities that the new restarter component must have.
+
+#### Support for limited workload types
+
+After the analysis of current implementation it's been identified that support for Pods and Jobs was not implemented.
+Instead, the restarter was returning a Warning.
+With ReplicaSet not owned by anything, we delete pod and hope replica controller would heal it.
```
mluk-sap: But what is wrong with that?
We delete pods, because we expect that new pods would come up with a new istio-proxy version.
Similarly, we annotate deployment, because we expect that pods would be replaced by ones with new-proxy version.
Both approaches are relying on a side-effect (webhook injecting new proxy version), as there isn't any better way of performing an upgrade, as its logic is fully managed in a webhook.

docs/contributor/adr/0018-restarter-component.md
mluk-sap: My previous comment applies here as well: https://github.com/kyma-project/istio/pull/2144/changes#r3702404010

AFAIK, we differentiate the case of Replicaset being owned by a deployment from ReplicaSet without an owner.

Having a ReplicaSet with a single replica sounds like a misconfiguration... We could treat it similarly to bare pod - report a warning, etc.

docs/contributor/adr/0018-restarter-component.md
mluk-sap: This description is a little bit unclear.

I understand the idea of Predicates as some piece of code that decides whether to restart 'something' or not. This 'Something' is passed as argument, decision is returned as boolean. Predicate 'instance' is a singleton - constructed once and reused for all further restarts during restarter component lifespan.

Other information (like expected proxy version) should be passed in a different way:
- via constructor
- at runtime (Predicate called directly (outside of interface)
- predicate itself may get it in an own way (api call, etc.)

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,483 @@
...
+- Exponential memory growth due to the number of API calls to the Kubernetes API server, which can lead to performance
+  degradation.
+- Execution during the Istio CR reconciliation, which can cause delays in applying configuration changes and affect the
+  overall responsiveness of the application.
+- Lack of support for customizability for customer to configure and control the restart behavior of workloads, which can
+  lead to unexpected restarts and potential downtime.
+- Code quality issues that can make it difficult to maintain and extend the restarter component, leading to increased
+  technical debt and bugs.
+
+### Background
```
mluk-sap, line 44: There is also more issues here:
- infinite restarts in case of failures
- lot of imperative code that is not unit testable
- predicates always restarting the whole data plane as they don't know whether the given Istio CR state was already applied

docs/contributor/adr/0018-restarter-component.md
mluk-sap, line 1: This ADR doesn't explain how to solve the race condition:
- Istio controller is being upgraded -> it upgrades the Istiod with its webhook -> new Istiod comes up -> old Istiod is being stopped - this is a very important point in time when new webhook is being effective
- Restarted component is being upgraded/restarted

So the restarter component doesn't know when new webhook brings new istio version (and potentially even some new behavior).

mluk-sap, line 1: It would be nice to also solve the very big issue that we have with some predicates that can't check whether the given configuration change was already applied.
Currently it causes (potentially) unnecessary restarts of the whole data plane.
We may solve this with annotations / labels.
For example:
instead of 
```
restarted-at: <timestamp>
```
we could have sth like:
```
restarted-because-of-istio-version: 1.28.3
restarted-because-of-istio-cr-hash: ab1234ef (counted from relevant IstioCR attributes)
```

With this approach we have an information about the restart reason, so we can restart the workload only once.

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,483 @@
...
+
+We have decided to implement a new restarter component that will address the limitations stated above. The new component
+will be designed to be more efficient, reliable, and customizable, while also improving code quality and
+maintainability.
+The new restarter component will be implemented as a separate deployment that will run independently of the Istio
+reconciliation loop, but the lifecycle of it will be managed by the Istio module.
+The component will be responsible for monitoring the workloads in the Istio service mesh and restarting them as needed,
+based on the default configuration and overrides provided by the user.
+
+Because the new restarter will work on the asynchronous event-driven model, it will fetch the list of Pods from
```
mluk-sap: This is not clear, my understanding is that it won't be event-driven.
So the new restarter can't even know when Istio-proxy with a new webhook comes up.

docs/contributor/adr/0018-restarter-component.md
mluk-sap, line 1: We should also think how to avoid infinite restarts.  This happened multiple times already. The good example is native sidecar predicate, which was expecting the native sidecar to be enabled, while the webhook did not enable it (because of old kubernetes version).

The approach with annotations / labels like in the previous comment would probably solve this.

mluk-sap, line 1: We should also describe that the restarter is split into two separate areas:
- restart of directly managed resources (like Ingresses)
  - will be even more valid with the Gateway API
- restart of the whole data plane (incl. customer workloads)

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,483 @@
...
+"top-to-bottom". This updated behavior starts from the parent workload (Deployment, StatefulSet, DaemonSet) and then
+uses selectors to retrieve all child Pods to admit.
+This will greatly reduce the number of API calls and sizes of the objects to admit in single loop. It will also ensure
+that admitted workload is up-to-date with the current cache state.
+
+### Lifecycle of the component
```
mluk-sap: It was confusing here, in fact it is a lifecycle of the **restarter** component

docs/contributor/adr/0018-restarter-component.md
Ressetkk: Predicates should be initialized on each reconciliation loop, though the list of them (incl. constructors) should rather be static. Singleton makes no sense, as we must evaluate the *current* state of the object, and with many concurrent loops, we'd have a lot of race conditions. It's the similar approach to what we currently have.

docs/contributor/adr/0018-restarter-component.md
Ressetkk: Deleting pods is unsafe. Simple as that. If you remove a pod from replicaset, and user currently is connected to it, you will crash their query and force 500 error. Even better, if the application covered by RS does some long-running task, you can crash it. If you rollout-restart you **wait** for new replicas to be ready, then redirect traffic and as a **final** step remove old replicas.

docs/contributor/adr/0018-restarter-component.md
mluk-sap: Okay, maybe I did not get the idea.

This is how I understood it:

Let's imagine sth like IstioProxyVersionMatcher:
- we construct a single instance of IstioProxyVersionMatcher and it is added to predicate list
- expected Istio proxy version is passed via constructor
- Matcher method is called separately for every workload object (Deployment, Statefulset, Replicaset instances)

Of course, we may initialize this list of predicates in every reconciliation loop, but by the 'singleton' I meant that the same predicate instance is used for multiple workloads.

docs/contributor/adr/0018-restarter-component.md
barchw: > Deleting pods is unsafe. Simple as that. If you remove a pod from replicaset, and user currently is connected to it, you will crash their query and force 500 error. Even better, if the application covered by RS does some long-running task, you can crash it. If you rollout-restart you wait for new replicas to be ready, then redirect traffic and as a final step remove old replicas.

That's not true at all. That's exactly what graceful termination is for. Rollout restart changes nothing in case of already existing connections.

docs/contributor/adr/0018-restarter-component.md
Ressetkk: Generally when I explored this topic last year I noticed that other implementations do not work on pods or other workloads not supporting rollout restarts. Only returned a notification that workloads needs restart.

docs/contributor/adr/0018-restarter-component.md
Ressetkk: We should check the adoption of the replicasets without owners on runtimes then and make decision wether we want to support it in the first iteration or not.

docs/contributor/adr/0018-restarter-component.md
Ressetkk: I think it would be best to leave it as part of the implementation. IMO the last-applied-configuration we have is also fine. We can also agree that the predicates *may* also project some additional metadata *in the workload* with the checksum for each option to keep the source of truth.

Ressetkk: When Istio components are getting up the restarter still works and reconciles the workloads against the older version of istio. It's actually fine because at the time of istio upgrade, usually old webhook would only be available.

Restarter is updated after new istio is up and running. If anytime between istio being ready and before restarter upgrade, it's possible to have an eventual workload restart, but that would be very small margin since upgrade of restarter will trigger reconciliation of all workloads and evaluate them to restart. This approach ensures that all workloads are **eventually up-to-date**, with periodic checks and event-based model. If not immediately, then after particular period of time (15m default).

Ressetkk: I think this must be testable on implementation level, as part of implementation requirements. The example you gave was a bug in a predicate logic. Any kind of safeguard whereas is always better, we should rather focus to improve on the test level, rather than introducing stateful "safeguard", which might be tricky with request-based stateless reconciliation.

A unit/integration test should be enough to test each predicate:
- evaluates a workload and triggers restart
- controller updates the workload
- another evaluation does not trigger restart

Funnily, we already have some behavioral E2E tests that test this particular case.

Ressetkk: Do we want to implement this separation in new component?

kubicar (dismissed):

docs/contributor/adr/0018-restarter-component.md
mluk-sap: My proposal of having the istio version in the annotation makes sense if we know that the new istiod with new webhook is already effective.
The restarter component would be even more separated from the Istio operator, so we don't really know what is being injected by the webhook. Thus, looking at Istio-proxy version is probably more reliable and simpler.

I forgot about the last-applied-configuration annotation. This fulfills the requirement of restarting pods only once.

But the cr-hash idea has one more advantage - the hash may be counted only from attributes that do matter in a given case. For example changing the numTrustedProxies parameter influences only the Ingresses, so we don't need to restart the whole mesh.

So maybe we should propose that in this spec as a potential improvement (we don't need this in MVP)

mluk-sap: We have a chicken-or-egg problem here. At day 0 we don't know that we have a bug in the implementation, so consequently we won't write test for it. If we would have predicted that then we had no issue at all.

So I'm mainly talking about the non-obvious cases which are not testable.

So for example - the predicate that was checking for native sidecars did not have a condition about the Kubernetes version, because we did not know this limitation, it was never documented and it hasn't been detected in dev/stage environments as all clusters were correctly upgraded there.

The impact of this misbehavior was quite huge, it was infinitely restarting all workloads in affected clusters.

So having some kind of circuit breaker in this mechanism would limit the blast radius of this bug.

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -221,6 +239,36 @@ This mode excludes workload from **all restarts**, including Istio updates that
 Using this annotation also implies that the user is aware of the risks of not restarting the workload and take a
 responsibility to keep it updated.
 
+The module will emit a `RestartRequired` event for excluded workloads that must be restarted, so the user can
+take action to restart them manually.
+
+#### Watching namespace and workload-level istio injection changes
+
```
barchw, line 243: This behaviour is entirely new: we never automatically enrolled user workloads to the mesh in case they switched NS label

docs/contributor/adr/0018-restarter-component.md
barchw, line 1: It's still missing the info of what happens on Istio CR edit (changing setting meaningful for restarts). Is it:
- a reconcile is triggered for all resources with EnqueueRequest?
- nothing?
- something else?

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -372,44 +470,33 @@ invalid annotation does not spam the Events API.
...
+control plane — a real `kube-apiserver` and `etcd`, no kubelet. They assert the reconciliation path end-to-end at the
+controller level: mutate a watched field, then check the reconciler enqueues, evaluates the rule chain, and issues the
+expected restart patch. They also cover what unit tests cannot reach — event emission, requeue/rate-limiting, and that
+the cache transform retains every field the rules read. With no kubelet, pods are simulated, so this tier validates
+control flow rather than real rollouts.
+
+**E2E tests** deploy the restarter on a live cluster alongside a real Istio installation and validate the
+customer-observed outcome: apply a config change, then assert targeted workloads are rolled and become ready while
+excluded and maintenance-window workloads are untouched. This is the only tier that verifies real pod rollouts and the
+full module lifecycle (install, upgrade, uninstall). Being slow and brittle, it covers critical customer journeys only.
```
barchw, line 475: 1. "critical customer journeys only" is rather not well defined.
2. The E2E test concept is acceptable, but generally risky.

barchw (commented):
I won't block the PR, but I would rather get an answer to https://github.com/kyma-project/istio/pull/2144/changes/f0132dc2cb2bddd0470c650dfa81cdac079adc91#r3753264053

docs/contributor/adr/0018-restarter-component.md
Ressetkk: It was something that @mluk-sap and @kubicar proposed. We agreed to add it for the record.

docs/contributor/adr/0018-restarter-component.md
Ressetkk: Current idea was to do nothing, and wait for the restarter to **eventually** reconcile the mesh over the `reconcileInterval`, unless we modify the configuration of the restarter itself.

barchw: Do you think it would be bad to perform EnqueueRequest? The requests would get queued and we can utilize controller runtime concurency, getting a gradual mesh rollout out of the box

barchw: Like this (default MaxConcurrentReconciles is 1):
```go
WithOptions(controller.Options{
  MaxConcurrentReconciles: 10,
}).
```

mluk-sap: I think we raised it also with @kubicar that the user expectation is that the controller applies changes quickly.  I would say that the max acceptable time is ~1 minute, for sure the user can't wait for let's say 10 minutes to see any effect.
So if we can speed it up by such request then I'm for it 👍

Ressetkk: Yeah. It's a good thing to have. I'll extend the docu then.

mluk-sap: No we don't need this. I somehow assumed that this will be still separated...

docs/contributor/adr/0018-restarter-component.md
mluk-sap: FYI we had a discussion with @Ressetkk who explained that helm is probably the simplest solution and this approach is used also in other modules like serverless (similar use case)

barchw (dismissed):

docs/contributor/adr/0018-restarter-component.md
mluk-sap: Yes, to confirm. The idea was to predict such functionality, so our design is prepared for it.

It doesn't mean that we'll implement it during the refactor itself. It can be added as a separate feature.

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,535 @@
...
+Complete(r)
+```
+
+The reconcile logic itself is unchanged: whether triggered by a workload event, a namespace event, or the time-based
+requeue, it re-evaluates the same rule chain against the current workload state. The namespace watch only controls
+*when* a reconcile is enqueued, not *how* the decision is made.
+
+#### Extending the configuration and future-proofing
+
+These options are defined at the workload level. If there is a need to define them at the namespace level, we can
```
mluk-sap: Not clear which options. It probably refers to the exclusions...?

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,535 @@
...
+ReplicaSets, however, do not support rolling restarts. To restart workloads covered by a ReplicaSet, the restarter
+deletes the pods that belong to the ReplicaSet and relies on the ReplicaSet controller to create new ones.
+This special feature only supports ReplicaSets that are **not** managed by a parent object.
+This is a fragile approach, as it can cause downtime if the ReplicaSet does not have enough replicas to maintain
+availability or is not covered by a PodDisruptionBudget. The user must be aware of this and take responsibility for the
+risk of downtime.
+
+As mentioned in the Decision section, the component goes top-to-bottom. The controller first admits the parent object
+(Deployment, StatefulSet, DaemonSet, ReplicaSet) and then uses selectors to retrieve all child Pods to admit in a single
+reconcile loop. This implies that some predicates are evaluated only on the parent object and some only on the child
```
mluk-sap: This is unclear for me. From one side we say that we operate on parents, but then we say that predicates may operate on both levels

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,535 @@
...
+#### Restart exclusions
+
+The component must support a configuration that allows the user to specify workloads that should be excluded from
+restarts.
+This helps to avoid restarts of critical workloads that cannot tolerate downtime. We introduce a new annotation
+`restarter.istio-operator.kyma-project.io/exclude` that supports a boolean value of `true` or `false`. If the annotation
+is set to `true`, the workload is excluded from restarts.
+If the annotation is not present, the workload is included in restarts by default.
+
+This mode excludes the workload from **all restarts**, including Istio updates that mitigate a CVE in the proxy.
```
mluk-sap: We should make it clear that it excludes the given workload only from our restarter, while the workload can still be restarted because of other events like Kubernetes upgrade.
Also, let's make it clear that it also doesn't cover things like Istio component restarts, which are result of Istio installation.

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,535 @@
...
+
+controller-runtime handles this with a secondary watch that maps the changed object back to the workloads it affects.
+Alongside the primary `For(&appsv1.Deployment{})` watch, the controller adds a `Watches` on `&corev1.Namespace{}` with
+an `EnqueueRequestsFromMapFunc` that lists the workloads in the changed namespace and enqueues a reconcile request for
+each. A `predicate` narrows namespace events to those where the injection label actually changed, so unrelated namespace
+edits do not trigger churn.
+
+```go
+ctrl.NewControllerManagedBy(mgr).
+For(&appsv1.Deployment{}).
```
mluk-sap: indentation missing

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,535 @@
...
+- **`Continue`** — this predicate is indifferent; the decision is deferred to the remaining predicates. This is the
+  neutral element of the default OR evaluation.
+- **`Stop`** — this predicate forbids the restart. Evaluation stops immediately and the workload is **not** restarted,
+  regardless of what any other predicate would decide.
+
+```go
+type Decision int
+
+const (
+Continue Decision = iota
```
mluk-sap: indentation missing

docs/contributor/adr/0018-restarter-component.md
mluk-sap: Okay, after discussions with @Ressetkk and @kubicar we agreed that we won't solve this for now.

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,537 @@
...
+Each predicate implements a single `Rule` interface with a single method `Evaluate()`.
+A workload is a Kubernetes object that implements the `Object` interface.
+Instead of returning a boolean, `Evaluate()` returns a tri-state `Decision`. A boolean can only express "restart" or
+"don't care", which forces conditions that *forbid* a restart (such as a maintenance window or an exclusion annotation)
+to be handled outside the predicate list. A third state lets a predicate veto the restart from within the same chain,
+short-circuiting evaluation.
+
+The three possible decisions are:
+
+- **`Restart`** — the workload has drifted from its desired state and, unless vetoed, should be restarted.
```
mluk-sap: Should be:
Evaluation stops immediately and the workload is restarted

mluk-sap (dismissed):

kubicar (dismissed):

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,537 @@
...
+We have also identified that the current implementation causes the Istio module to have exponential memory growth on
+large clusters. The root cause is the number of API calls to fetch the list of workloads running in the Istio service
+mesh. To admit a workload for restart, be it a Deployment, StatefulSet or DaemonSet, the restarter works on a list of
+uncorrelated pods.
+In a single reconciliation loop, the restarter fetches the list of all pods in a cluster in paginated chunks of 100.
+For each chunk, the restarter admits the pods to the predicates, which usually requires additional API calls to retrieve
+the parent of an admitted pod during restart. Listing all pods in a cluster in a single reconcile loop, even with
+pagination, costs a lot of memory. On larger clusters we might end up with a stale cache, which can cause unwanted API
+errors and another restart run. With regular errors, the Istio module can end up in a loop of restarts, causing a lot of
+downtime for the customer.
```
nataliasitko, line 62:
```suggestion
Restarts are handled in one big bang because the component works within a single reconciliation loop of the Istio CR.
Restarting everything in a single run can cause massive downtimes that the customer does not expect.
We have also identified that the current implementation causes the Istio module to have exponential memory growth on
large clusters. The root cause is the number of API calls to fetch the list of workloads running in the Istio service
mesh. To admit a workload for restart, be it a Deployment, StatefulSet, or DaemonSet, the restarter works on a list of
uncorrelated Pods.
In a single reconciliation loop, the restarter fetches the list of all Pods in a cluster in paginated chunks of 100.
For each chunk, the restarter admits the Pods to the predicates, which usually requires additional API calls to retrieve
the parent of an admitted Pod during restart. Listing all Pods in a cluster in a single reconcile loop, even with
pagination, costs a lot of memory. On larger clusters, we might end up with a stale cache, which can cause unwanted API
errors and another restart run. With regular errors, the Istio module can end up in a loop of restarts, causing significant
downtime for the customer.
```

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,537 @@
...
+The new restarter works on an asynchronous, event-driven model. It watches workloads, reacts to changes, and reconciles
+them asynchronously, implementing the Kubernetes controller pattern. Additionally, it uses time-based reconciliation to
+regularly ensure that workloads are up to date with the current state.
+
+It fetches the list of Pods top-to-bottom. This updated behavior starts from the parent workload (Deployment,
+StatefulSet, DaemonSet) and then uses selectors to retrieve all child Pods to admit.
+This greatly reduces the number of API calls and the size of the objects to admit in a single loop. It also ensures that
+the admitted workload is up to date with the current cache state.
+
+### Lifecycle of the restarter component
```
nataliasitko:
```suggestion
### Lifecycle of the Restarter Component
```

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,537 @@
...
+3. **Drift detection**: The reconciler compares the desired chart values derived from the Istio CR against the
+   last-applied values stored in the Helm release secret. If they differ, an upgrade is triggered.
+
+The Helm release is stored in the same namespace as the Istio CR (usually `kyma-system`) as a Secret with Helm labels.
+This allows standard `helm list` introspection for debugging.
+
+Since the Istio version is tightly coupled to the module version, the restarter component version is also coupled to the
+Istio module version. This means that when the Istio module is updated, the restarter component is updated as well.
+
+Configuration of the restarter component is provided via a `.spec.restarter` field in the Istio CR. The reconciler
```
nataliasitko:
```suggestion
Configuration of the restarter component is provided via a **.spec.restarter** field in the Istio CR. The reconciler
```

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,537 @@
...
+          cpu: 200m
+          memory: 256Mi
+```
+
+Changing the lifecycle of the component also requires adjusting the decisions made
+in [ADR-0002](./0002-istio-cr-status-improvements.md) regarding the restart conditions.
+This change **drops** the `ProxySidecarRestartSucceeded` condition and instead introduces a new condition
+`RestarterComponentReady` that reflects the state of the restarter component. The new condition is set to `True` when
+the component is installed and running successfully, and `False` when it is not.
+The Istio module does not attempt to restart workloads if the restarter component is not ready, and logs a Processing
```
nataliasitko:
```suggestion
The Istio module does not attempt to restart workloads if the restarter component is not ready, and logs a `Processing`
```

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,537 @@
...
+| ~~Processing~~ | ~~ProxySidecarRestartSucceeded~~ | ~~True~~  | ~~ProxySidecarRestartSucceeded~~      | ~~Proxy sidecar restart succeeded~~                             |
+| ~~Error~~      | ~~ProxySidecarRestartSucceeded~~ | ~~False~~ | ~~ProxySidecarRestartFailed~~         | ~~Proxy sidecar restart failed~~                                |
+| ~~Processing~~ | ~~ProxySidecarRestartSucceeded~~ | ~~False~~ | ~~ProxySidecarPartiallySucceeded~~    | ~~Proxy sidecar restart partially succeeded~~                   |
+| ~~Warning~~    | ~~ProxySidecarRestartSucceeded~~ | ~~False~~ | ~~ProxySidecarManualRestartRequired~~ | ~~Proxy sidecar manual restart is required for some workloads~~ |
+| Ready          | RestarterComponentReady          | True      | RestarterComponentReady               | Restarter component is installed and running successfully       |
+| Processing     | RestarterComponentReady          | False     | RestarterComponentNotReady            | Restarter component is not ready                                |
+
+### Features and capabilities of the component
```
nataliasitko:
```suggestion
### Features and Capabilities of the Component
```

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,537 @@
...
+
+We have identified the following features and capabilities that the new restarter component must have.
+
+#### Support for limited workload types
```
nataliasitko:
```suggestion
#### Support for Limited Workload Types
```

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,537 @@
...
+
+After analyzing the current implementation, we identified that support for Pods and Jobs was not implemented. Instead,
+the restarter returned a Warning.
```
nataliasitko:
```suggestion
the restarter returned a warning.
```

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,537 @@
...
+This behavior is misleading, as it gives the impression that restarting these workloads is supported, while in reality
+they are not.
```
nataliasitko:
```suggestion
This behavior is misleading, as it gives the impression that restarting these workloads is supported, but in reality, it is not.
```

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,537 @@
...
+
+To further simplify the decision, the component supports restart only for workloads managed by a Deployment,
+StatefulSet, DaemonSet or ReplicaSet — that is, workloads that support rolling restarts or self-healing.
+To restart a workload, we use behavior similar to the `kubectl rollout restart` command, which updates the pod template
```
nataliasitko:
```suggestion
To restart a workload, we use behavior similar to the `kubectl rollout restart` command, which updates the Pod template
```

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,537 @@
...
+ReplicaSets, however, do not support rolling restarts. To restart workloads covered by a ReplicaSet, the restarter
+deletes the pods that belong to the ReplicaSet and relies on the ReplicaSet controller to create new ones.
+This special feature only supports ReplicaSets that are **not** managed by a parent object.
+This is a fragile approach, as it can cause downtime if the ReplicaSet does not have enough replicas to maintain
+availability or is not covered by a PodDisruptionBudget. The user must be aware of this and take responsibility for the
+risk of downtime.
+
+As mentioned in the Decision section, the component goes top-to-bottom. The controller first collects the parent object
+(Deployment, StatefulSet, DaemonSet, ReplicaSet) and then uses selectors to retrieve all child Pods to admit in a single
+reconcile loop. Then each workload (Parent + children) is evaluated through a set of predicates in a loop. If predicate
```
nataliasitko:
```suggestion
reconcile loop. Then each workload (parent and children) is evaluated through a set of predicates in a loop. If a predicate 
```

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,537 @@
...
+As mentioned in the Decision section, the component goes top-to-bottom. The controller first collects the parent object
+(Deployment, StatefulSet, DaemonSet, ReplicaSet) and then uses selectors to retrieve all child Pods to admit in a single
+reconcile loop. Then each workload (Parent + children) is evaluated through a set of predicates in a loop. If predicate 
+supports only one type of workload, it must implement a type assertion and skip the evaluation for unsupported workload
+types.
+
+If demand for support of other workload types arises, we can implement it in the future. However, we do not support
+other workload types in the first implementation.
+
+#### Maintenance window support
```
nataliasitko:
```suggestion
#### Maintenance Window Support
```

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,537 @@
...
+
+Validation parses the annotation and rejects invalid syntax by emitting an appropriate Event.
+The check at reconcile time becomes:
+
+1. Parse the day range → is `time.Now().Weekday()` within it?
+2. Parse the time range → is the current time-of-day within it?
+3. If the maintenance window is smaller than `reconcileInterval`, emit a `RestartConfigInvalid` event and skip.
+4. Both true → proceed with restart; otherwise skip and requeue with `reconcileInterval` until the window opens.
+
+#### Restart exclusions
```
nataliasitko:
```suggestion
#### Restart Exclusions
```

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,537 @@
...
+
+This mode excludes the workload from **restarts handled by component**, including Istio updates that mitigate a CVE in 
+the proxy. This does **NOT** exclude the workload from restarts caused by other means, like node drains.
+Using this annotation also implies that the user is aware of the risk of not restarting the workload and takes
+responsibility for keeping it updated.
+
+The module emits a `RestartRequired` event for excluded workloads that must be restarted, so the user can take action to
+restart them manually.
+
+#### Watching namespace and workload-level Istio injection changes
```
nataliasitko:
```suggestion
#### Watching Namespace and Workload-Level Istio Injection Changes
```

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,537 @@
...
+        builder.WithPredicates(injectionLabelChanged()),
+    ).
+Complete(r)
+```
+
+The reconcile logic itself is unchanged: whether triggered by a workload event, a namespace event, or the time-based
+requeue, it re-evaluates the same rule chain against the current workload state. The namespace watch only controls
+*when* a reconcile is enqueued, not *how* the decision is made.
+
+#### Extending the configuration and future-proofing
```
nataliasitko:
```suggestion
#### Extending the Configuration and Future-Proofing
```

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,537 @@
...
+
+The options above (maintenance window, restart exclusion) are defined at the workload level. If there is a need to 
+define them at the namespace level, we can introduce it in the future as a separate custom resource and plan a migration
+from workload-level annotations to a custom resource with label selectors.
+This allows configuring restarter behavior for multiple workloads with a single configuration.
+
+### Predicate logic for workload restart
```
nataliasitko:
```suggestion
### Predicate Logic for Workload Restart
```

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,537 @@
...
+front of the predicate list means a workload inside a maintenance window or explicitly excluded is never restarted, no
+matter how many drift predicates would otherwise match.
+
+The predicates must be extensible so that each predicate can be used in different reconciliation loops.
+The list of available predicates must be defined in a single place within a reconcile loop, as a list of `Rule`
+implementations.
+This list cannot mutate, meaning predicates cannot be added or removed at runtime. This ensures that the predicates are
+evaluated in a consistent order and that the behavior of the component is predictable.
+
+#### Naming scheme
```
nataliasitko:
```suggestion
#### Naming Scheme
```

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,537 @@
...
+
+- Veto rules that return `Stop` follow the same principle: name the active condition that causes the veto, not the veto
+  itself. Examples: `MaintenanceWindowActive` (a window is currently open) and `WorkloadExcluded` (the workload carries
+  an exclusion annotation).
+
+These forms keep the name focused on *why* rather than *what*. A rule name must not describe the restart itself (for
+example, `RestartOnCni` or `DoProxyRestart` are discouraged) — the restart is implied by an `Restart` decision, and the
+veto by a `Stop`.
+
+#### Logical operations of predicates
```
nataliasitko:
```suggestion
#### Logical Operations of Predicates
```

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,537 @@
...
+- if every rule returns `Continue`, the workload is left untouched.
+
+The priority of predicates is defined by the actions they implement. When implementing a new predicate, the developer
+must ensure that the predicate is added to the list in the correct order, so that it is evaluated in the right context.
+The order of predicates is defined by the actions they implement:
+
+- Predicates that return `Stop` must be evaluated first, as they veto the restart and short-circuit evaluation;
+- Predicates that return `Restart` must be evaluated after the `Stop` predicates.
+
+### Event handling and reconciliation process
```
nataliasitko:
```suggestion
### Event Handling and Reconciliation Process
```

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,537 @@
...
+
+For better stability of the restarter, the implementation follows the standard Kubernetes controller pattern.
+
+#### Manager and controllers
```
nataliasitko:
```suggestion
#### Manager and Controllers
```

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,537 @@
...
+server.
+
+The restarter component watches for events on the supported workload types (Deployment, StatefulSet, DaemonSet,
+ReplicaSet). Each workload type is covered by a controller. Controllers define watches for the supported workloads. When
+an event is received, the controller enqueues a reconcile request for the workload. The events are asynchronous, which
+ensures that workloads are not admitted at the same time. Additionally, the restarter implements a time-based requeue
+mechanism to ensure that workloads are regularly checked for drift, even if no events are received. This is done by
+setting a `reconcileInterval` in the configuration, which triggers a reconcile request for each workload.
+
+For mesh configuration changes, e.g. trusted proxies or Prometheus merge changes, to ensure that customers get the
```
nataliasitko:
```suggestion
For mesh configuration changes, for example, trusted proxies or Prometheus merge changes, to ensure that customers get the
```

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,537 @@
...
+token-bucket limiter, which caps the queue's aggregate throughput. A token-bucket limiter takes a sustained rate `r`
+(tokens per second) and a burst size `b` (bucket capacity), refilling the bucket at `r` tokens per second so the
+sustained average never exceeds `r`. This ceiling applies per controller; a single shared limiter can be passed to all
+controllers if a global cap across workload kinds is required.
+
+To tackle memory issues, the manager configures a cache transform on the watched workload types.
+controller-runtime's cache supports a `TransformFunc` that is applied to every object before it is stored in the
+informer cache. By setting a `TransformFunc` on each `ByObject` entry (or globally via `DefaultTransform`), the
+restarter can strip fields that are never used in predicate evaluation — such as `managedFields`, `ownerReferences`, and
+the full pod spec — before the object is committed to memory.
```
nataliasitko:
```suggestion
the full Pod spec — before the object is committed to memory.
```

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,537 @@
...
+expected restart patch. They also cover what unit tests cannot reach — event emission, requeue/rate-limiting, and that
+the cache transform retains every field the rules read. With no kubelet, pods are simulated, so this tier validates
+control flow rather than real rollouts.
+
+**E2E tests** deploy the restarter on a live cluster alongside a real Istio installation and validate the
+customer-observed outcome: apply a config change, then assert targeted workloads are rolled and become ready while
+excluded and maintenance-window workloads are untouched. This is the only tier that verifies real pod rollouts and the
+full module lifecycle (install, upgrade, uninstall). Being slow and brittle, it covers critical customer journeys only.
+
+### Metrics and observability
```
nataliasitko:
```suggestion
### Metrics and Observability
```

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,537 @@
...
+| ~~Processing~~ | ~~ProxySidecarRestartSucceeded~~ | ~~True~~  | ~~ProxySidecarRestartSucceeded~~      | ~~Proxy sidecar restart succeeded~~                             |
+| ~~Error~~      | ~~ProxySidecarRestartSucceeded~~ | ~~False~~ | ~~ProxySidecarRestartFailed~~         | ~~Proxy sidecar restart failed~~                                |
+| ~~Processing~~ | ~~ProxySidecarRestartSucceeded~~ | ~~False~~ | ~~ProxySidecarPartiallySucceeded~~    | ~~Proxy sidecar restart partially succeeded~~                   |
+| ~~Warning~~    | ~~ProxySidecarRestartSucceeded~~ | ~~False~~ | ~~ProxySidecarManualRestartRequired~~ | ~~Proxy sidecar manual restart is required for some workloads~~ |
+| Ready          | RestarterComponentReady          | True      | RestarterComponentReady               | Restarter component is installed and running successfully       |
+| Processing     | RestarterComponentReady          | False     | RestarterComponentNotReady            | Restarter component is not ready                                |
```
nataliasitko:
```suggestion
| `Processing` | `ProxySidecarRestartSucceeded` | `True`  | `ProxySidecarRestartSucceeded`      | Proxy sidecar restart succeeded                            |
| `Error`      | `ProxySidecarRestartSucceeded` | `False` | `ProxySidecarRestartFailed`         | Proxy sidecar restart failed                                |
| `Processing` | `ProxySidecarRestartSucceeded` | `False` | `ProxySidecarPartiallySucceeded`    | Proxy sidecar restart partially succeeded                   |
| `Warning`    | `ProxySidecarRestartSucceeded` | `False` | `ProxySidecarManualRestartRequired` | Proxy sidecar manual restart is required for some workloads |
| `Ready`          | `RestarterComponentReady`          | `True`      | `RestarterComponentReady`               | Restarter component is installed and running successfully       |
| `Processing`     | `RestarterComponentReady`          | `False`     | `RestarterComponentNotReady`            | Restarter component is not ready                                |
```

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,537 @@
...
+"don't care", which forces conditions that *forbid* a restart (such as a maintenance window or an exclusion annotation)
+to be handled outside the predicate list. A third state lets a predicate veto the restart from within the same chain,
+short-circuiting evaluation.
+
+The three possible decisions are:
+
+- **`Restart`** — the workload has drifted from its desired state and should be restarted.
+- **`Continue`** — this predicate is indifferent; the decision is deferred to the remaining predicates. This is the
+  neutral element of the default OR evaluation.
+- **`Stop`** — this predicate forbids the restart. Evaluation stops immediately and the workload is **not** restarted,
```
nataliasitko:
```suggestion
- `Restart` — the workload has drifted from its desired state and should be restarted.
- `Continue` — this predicate is indifferent; the decision is deferred to the remaining predicates. This is the
  neutral element of the default OR evaluation.
- `Stop` — this predicate forbids the restart. Evaluation stops immediately and the workload is **not** restarted,
```

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,537 @@
...
+This list cannot mutate, meaning predicates cannot be added or removed at runtime. This ensures that the predicates are
+evaluated in a consistent order and that the behavior of the component is predictable.
+
+#### Naming scheme
+
+Rule names must describe *the condition being asserted*, not the action taken. The name states the condition that the
+rule detects as **true**; the `Decision` it returns (`Restart` or `Stop`) determines the direction. A developer must be
+able to understand what the rule checks from its name alone, without knowing which decision it maps to.
+
+- **`<Subject>Changed`** — for rules that return `Restart` when the workload has drifted from its desired state. Read as
```
nataliasitko:
```suggestion
- `<Subject>Changed` — for rules that return `Restart` when the workload has drifted from its desired state. Read as
```

docs/contributor/adr/0018-restarter-component.md
```diff
@@ -0,0 +1,537 @@
+# Restarter component in Istio module
```
nataliasitko:
```suggestion
# Restarter Component in the Istio Module
```

nataliasitko (approved):

kubicar (approved):

PR #2178, opened 2026-08-13
proxy stats matcher

### LINKED ISSUE #1661: Configurability of proxyStatsMatcher for Kyma managed resources
**Description**

By default the Istio disables some metrics, like 'outlier_detection' or 'retry' ones. These metrics can be enabled:
- either globally in meshConfig
- or by 'proxy.istio.io/config' annotation on a given workload.

Example of such configuration:
```
      proxyStatsMatcher:
        inclusionRegexps:
        - ".*outlier_detection.*"
        - ".*upstream_rq_retry.*"
        - ".*upstream_cx_.*"
        inclusionSuffixes:
        - "upstream_rq_timeout"
```

See more details: https://istio.io/latest/docs/ops/configuration/telemetry/envoy-stats/

The problem is that metrics (like 'outlier_detection' or 'retry') needs to be applied and analyzed on the **source proxy** that makes a call to a given destination. So in order to analyze Ingress->service communication users need to apply these annotations on Istio Ingress, which a Kyma managed resource. This is not recommended and may not work properly (deleted/modified by reconciliation, etc.).

We need to offer some official way of applying these settings:
- either globally (Istio CR translated to the meshConfig)
- or for a managed service (like Ingress/Egress config in the Istio CR)
- or just 'officially' allow users to do it via annotations (document it, guarantee to not delete it in reconciliation, etc.)

**Reasons**

Users need this to diagnose problems. In case of Ingress->Service communication users should not be forced to touch Kyma-managed resources (like Ingress).

**Attachments**

kolodziejczak:
**Description**

- Extend module to allow configuration of the proxy stats matcher

**Pre-Merge Checklist**

- [x] The code coverage is acceptable.
- [x] Release notes for the introduced changes are created.
- [x] If Kubebuilder changes were made, you ran `make manifests` and committed the changes before the merge.
- [x] Pre-existing managed resources are correctly handled.
- [x] The change works on all hyperscalers supported by SAP BTP, Kyma runtime.
- [x] There is no upgrade downtime.
- [x] For infrastructure changes, you checked if the changes affect the hyperscaler's costs.
- [x] RBAC settings are as restrictive as possible.
- [x] If any new libraries are added, you verified license compliance and maintainability, and made a comment in the PR with details. We only allow stable releases to be included in the project.
- [x] You checked if this change should be cherry-picked to active release branches.
- [x] The configuration does not introduce any additional latency.
- [x] You checked if Busola updates are needed.
- [x] If you added a predicate for restarters, check if the **lastAppliedConfiguration** annotation is properly updated after the given restart is executed.

**Related issues**


internal/restarter/predicates/proxy_stats_matcher.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	return &ProxyStatsMatcherRestartPredicate{
+		oldInclusionRegexps: oldRegexps,
+		newInclusionRegexps: newRegexps,
+	}, nil
+}
+
+func (p ProxyStatsMatcherRestartPredicate) Matches(_ v1.Pod) bool {
+	return !reflect.DeepEqual(p.oldInclusionRegexps, p.newInclusionRegexps)
```
Ressetkk: When for some reason user initializes the `istioCR.Spec.Config.ProxyStatsMatcher.InclusionRegexps = []` from `istioCR.Spec.Config.ProxyStatsMatcher = nil`, this would trigger a restart, because `DeepEqual` treats `nil` and `[]string{}` as different. If that's not intended then it would be good to cover up this case and not trigger a restart when both slices are empty. `len()` returns 0 for `nil` and `[]string{}`.

Unless istio also treats them differently, and initialized `istioCR.Spec.Config.ProxyStatsMatcher.InclusionRegexps = []` is treated as "exclude all" than "use"default".

Ressetkk (changes requested):
I don't like it requires a new predicate, and triggering `GetLastAppliedConfiguration` inside a predicate is inherently bad design. We could think about moving this function outside predicates' constructors, call it before constructors and pass istio CR as an argument.

Besides one comment, everything LGTM.

Ressetkk (dismissed):

docs/user/04-00-istio-custom-resource.md
```diff
@@ -319,6 +320,17 @@ Appears in:
...
+### ProxyStatsMatcher
+
+Configures the stats matcher for Istio proxy sidecars and gateways.
+
+Appears in:
+- [Config](#config)
+
+| Field | Description | Validation |
+| --- | --- | --- |
+| **inclusionRegexps** <br /> string array | Defines a list of regular expressions used to select proxy stats for inclusion.<br />Stats whose names match any of these regular expressions are included in the proxy stats output. | Optional <br /> |
```
nataliasitko:
```suggestion
| **inclusionRegexps** <br /> string array | Defines a list of regular expressions used to select proxy statistics for inclusion.<br />Statistics with names that match any of these regular expressions are included in the proxy statistics output. | Optional <br /> |
```

Ressetkk (dismissed):

docs/release-notes/1.30.0.md
```diff
@@ -0,0 +1,4 @@
+
+## Enhancements
+
+- Added support for `proxyStatsMatcher` in the Istio CR, allowing you to configure which proxy stats are emitted by Istio sidecars and gateways using inclusion regular expressions. See [#2178](https://github.com/kyma-project/istio/pull/2178)
```
nataliasitko:
```suggestion
- You can now configure **proxyStatsMatcher** in the Istio custom resource. It allows you to define which proxy statistics are emitted by Istio sidecars and gateways using inclusion regular expressions. See [#2178](https://github.com/kyma-project/istio/pull/2178).
```

nataliasitko (approved):

Ressetkk (dismissed):

nataliasitko (dismissed):

nataliasitko (approved):

Ressetkk (approved):

